PR #72, opened 2020-11-25
Nested api field

pmbohm:


ui/src/components/data-entities/form-reducer.js
```diff
@@ -33,6 +34,21 @@ const formSlice = createSlice({
...
+      state.editItem = {...state.editItem, ...resp};
+    },
+    selectedItemsLoaded: (state, action) => {
+      let resp = {};
+      const key = Object.keys(action.payload._embedded)[0];
+      resp[key] = action.payload._embedded;
+      resp[pluralize.singular(key) + "Selected"] = action.payload.selected;
+      resp[pluralize.singular(key)] = action.payload.selected._links.self.href;
+      state.editItem = {...state.editItem, ...resp};
+    },
```
utas-tajona, line 51: I have pull the branch locally and it's crashing here. I don't see where action.payload.selected get fill ?

Especially in the case of new element.

ui/src/components/data-entities/customWidgetFields/NestedApiField.js
```diff
@@ -0,0 +1,56 @@
...
+import {Typography} from "@material-ui/core";
+import {selectedItemsRequested, setNestedField} from "../middleware/entities";
+import {markupProjectionQuery} from "../../utils/helpers";
+
+
+const NestedApiField = (props) => {
+
+  let editItemValues = useSelector(state => state.form.editItem);
+  const dispatch = useDispatch();
+
```
utas-tajona, line 17: the props has the currently being edited

ui/src/components/data-entities/customTooltip.js
```diff
@@ -0,0 +1,37 @@
...
+            </div>
+        );
+      }
+    }
+
+    return (<div className={"custom-tooltip ag-tooltip-hiding"} ></div>)
+
+
+  }
+}
```
utas-tajona, line 37: I m not sure we need to add a custom tooltip. ag-grid knows how to deal with large text: https://blog.ag-grid.com/wrapping-column-header-text/

ui/src/components/data-entities/customWidgetFields/NestedApiField.js
```diff
@@ -0,0 +1,56 @@
...
+  const entity = props.name ;
+  const pluralEntity = pluralize(entity);
+  //const entities =  pluralEntity.charAt(0).toLowerCase() + pluralEntity.slice(1);
+
+  let itemsList = (editItemValues[pluralEntity]) ?editItemValues[pluralEntity][pluralEntity] : [];
+  let selectedItems = (editItemValues[entity + "Selected"]) ? [editItemValues[entity + "Selected"]].filter(Boolean) : [];
+
```
utas-tajona, line 24: this need to be verified if we need to deal with multiple choice. IMO we don't

ui/src/components/data-entities/customWidgetFields/NestedApiField.js
```diff
@@ -0,0 +1,56 @@
...
+
+  useEffect(() => {
+    let urls = [pluralEntity + "?projection=selection"];
```
utas-tajona, line 27: this should be defined in the axios api

utas-tajona (changes requested):

ui/src/components/data-entities/customTooltip.js
pmbohm: I dont want to effect heights here

ui/src/components/data-entities/customWidgetFields/NestedApiField.js
pmbohm: The code can handle at no cost

ui/src/components/data-entities/customWidgetFields/NestedApiField.js
pmbohm: ok yes

pmbohm: moved to helpers.js

ui/src/ag-grid.scss
```diff
@@ -0,0 +1,56 @@
...
+}
+
+.custom-tooltip .body {
+  margin-left: 10px;
+  padding: 5px;
+  white-space: nowrap;
+  border: 2px solid grey;
+  background-color: white;
+  border-radius: 4px;
+}
```
utas-tajona, line 56: ag grid should follow the style of material . you can move the the tooltip style within its component

pmbohm: This is the right place for rules applying to AgGrid 
The rules are for Aggrid classes
Here's an example in their documentation: https://www.ag-grid.com/javascript-charts-tooltips/#styling-the-default-tooltip

pmbohm:
@utas-tajona Bug found and squashed

PR #74, opened 2020-11-25
Add an endpoint that ingests from a validated job

akashisama:
For https://github.com/aodn/backlog/issues/2357

Just lacking integration tests for the controller. The validation code isn't setting measures yet so my progress on that was stalled, but should be easy to finish off once the measures are in place (don't want to do any mocking since I really want to test that everything is working together [which of course it isn't right now due to the lack of measures])

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
@@ -1,0 +1,124 @@
+package au.org.aodn.nrmn.restapi.service;
+
+import au.org.aodn.nrmn.restapi.model.db.*;
+import au.org.aodn.nrmn.restapi.repository.*;
+import au.org.aodn.nrmn.restapi.validation.StagedRowFormatted;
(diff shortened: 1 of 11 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
```diff
@@ -0,0 +1,124 @@
...
+
+        int depth = Integer.parseInt(splitDepth[0]);
+        int survey_num = Integer.parseInt(splitDepth[1]);
+
+        Optional<Survey> existingSurvey = surveyRepository.findOne(Example.of(Survey.builder()
+                .depth(depth)
+                .surveyNum(survey_num)
+                .site(stagedRow.getSite())
+                .surveyDate(Date.valueOf(stagedRow.getDate()))
+                .build()));
```
akashisama, line 60: @jonescc ah, I am using examples in this PR, here's a findOne example

utas-tajona (approved):
LGTM but it will need some rework when the formatted add surveyNum

PR #75, opened 2020-11-26
Fixes aodn/backlog#2345 - disable concurrent branch builds

jonescc:
Apparently the workspace for a branch build is shared which causes all sorts of problems.  Thanks @lwgordonimos


lwgordonimos:
Seems to be behaving as expected now, note build 4 is pending because 3 has not completed yet:

image

PR #77, opened 2020-11-26
Adjust schema for associations to reflect what should be sent/mark them

jonescc:


utas-tajona (approved):
LGTM

jonescc:
So that doesn't work.  react form validates the request that its going to send against the schema.

Changing this to type = string format = uri

jonescc:
We can use an array of strings to reflect that multiple uris can be submitted.

jonescc:
Can now submit required uri to backend

utas-tajona (approved):
LGTM

PR #80, opened 2020-11-26
Add power user and data officer roles and default access

jonescc:


jonescc:
Fixed intermittent test failure for stagedjoblog and stagedjob.  create/modified times may equal recorded start or end times sometimes.

akashisama (approved):

PR #83, opened 2020-11-27
Make UserActionAudit constructor get auth details itself

akashisama:
Noticed that where the UserActionAudit is being called we've been using the constructor that doesn't take an Authentication object, which means that user isn't being stored. This change makes that constructor try to retrieve authentication from the context (```getContext()``` is guaranteed to never return null).


utas-tajona (approved):
LGTM

PR #84, opened 2020-11-30
Fix various build issues

jonescc:
* use jenkins global yarn cache to speed up builds
* leave node_modules alone when cleaning workspace 
* use more recent version of yarn
* lock package versions selected for consistent builds

### DIFF
--- a/ui/pom.xml
+++ b/ui/pom.xml
@@ -13,7 +13,7 @@
 
     <properties>
         <node.version>v12.18.3</node.version>
-        <yarn.version>v1.12.1</yarn.version>
+        <yarn.version>v1.22.10</yarn.version>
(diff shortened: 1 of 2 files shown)


ui/pom.xml
```diff
@@ -13,7 +13,7 @@
 
     <properties>
         <node.version>v12.18.3</node.version>
-        <yarn.version>v1.12.1</yarn.version>
+        <yarn.version>v1.22.5</yarn.version>
```
pmbohm, line 16: I have just put up a pull request to replace the deleted yarn.lock 
My version of yarn is currently 1.22.10

jonescc: OK updated

PR #85, opened 2020-11-30
Allow node_modules directory to be specified on the command line

jonescc:
```
mvn clean package -Dyarn.modules-folder=/tmp/node_modules
```


jonescc:
Its leads to a maze of twisty passages all alike

PR #86, opened 2020-12-01
Add basic read operations on staged job and staged job log

jonescc:
For use finding and selecting staged jobs

Also use consistent case for rest api tags


utas-tajona (approved):
LGTM

PR #87, opened 2020-12-01
Reinstate yarn.lock

pmbohm:
Yarn.lock was removed

We have previously agreed to use yarn for development and also for deployment builds


jonescc:
In the PR I just raised for you to review.

jonescc:
OK. Mine was first but happy to just get it in anyway I can.

PR #89, opened 2020-12-02
Feature/survey validation ui

utas-tajona:
refactoring so the validation api call can return:
Every row with their errrors
list of global errors
input errors // if the  job isn't found


jonescc (commented):
Looks good to me - some minor foirmatting/naming comments only

pmbohm (changes requested):
Linting changes should not be part of this pull request and will cause massive merge conflicts if merged

ccmoloney:
Is this ready for merging?  @utas-tajona

utas-tajona:
splited in different PR

PR #98, opened 2020-12-16
Fixes aodn/backlog#2503

jonescc:
Redirect list requests received by the backend to the react app for handling

We switched from collection to list for these pages at some point and the front-end whitelist was not updated


jonescc:
Should also fix https://github.com/aodn/backlog/issues/2513

pmbohm (approved):

PR #102, opened 2020-12-18
Fixups for database upgrade notes after viewing on github

jonescc:


Updating_Databases.md
```diff
@@ -2,7 +2,7 @@
 
 The ddl for the new NRMN database is currently maintained in two places
 
-* [NRMN/db-migrate/dd](https://github.com/aodn/NRMN/tree/master/db-migrate) - amalgamated ATRC/RLS
+* [NRMN/db-migrate/ddl](https://github.com/aodn/NRMN/tree/master/db-migrate) - amalgamated ATRC/RLS
```
akashisama, line 5: could also change the url to point to the ddl directory

jonescc: Actually the ddl directory doesn't include the endpoints which is probably why the ddl was truncated above - I was in the process of deleting it and got distracted.   I'll remove it.

akashisama (approved):

PR #107, opened 2021-01-11
Species csv

akashisama:
Work for https://github.com/aodn/backlog/issues/2523


utas-tajona (commented):
It looks good but I m not a big fan of If (... != null).

I know it's been already merge but I think we could change to Optional and get and Empty List when empty.

jonescc:
> I think we could change to Optional and get and Empty List when empty

Good idea - I like the idea of an empty list instead of null and it looks easy enough to do (https://stackoverflow.com/questions/51789957/how-to-set-default-value-of-array-to-empty-with-requestparam/51789958).  I'll have a look.

nspool:
Superseded by https://github.com/aodn/nrmn-application/pull/271. Closing.

PR #111, opened 2021-01-13
backend change

utas-tajona:


pmbohm:
@akashisama Would you be able to look at this one?

api/src/main/java/au/org/aodn/nrmn/restapi/validation/model/RowWithValidation.java
```diff
@@ -0,0 +1,17 @@
...
+import cyclops.data.tuple.Tuple2;
+import cyclops.function.Monoid;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public class RowWithValidation<T> {
+    Seq<StagedRow> rows;
+    Validated<StagedRowError, T> valid;
```
akashisama, line 16: Could these be private?

api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedRow.java
```diff
@@ -30,7 +31,7 @@
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
 
-    @JsonProperty(value = "Site No")
+    @JsonProperty(value = "site #")
```
akashisama, line 34: Wouldn't it be cleaner to just remove the ```@JsonProperty``` annotation completely? Same with site name and common name?

akashisama (changes requested):

api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedRowError.java
```diff
@@ -35,6 +36,11 @@
     @Column(name = "error_type")
     private ValidationCategory errorType;
 
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "error_level")
```
akashisama, line 41: This column will need to be added to the DDL as well right?

akashisama (changes requested):

api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedRowError.java
utas-tajona: StagedRowError is in a weird state. It's currently used as a regular model not JPA(never persited)
I have left the JPA definition in case we have  to save it for better performances once we get real(bigger) spreadSheet.

But I m happy to add the field to the DDL

api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedRow.java
utas-tajona: I think i can remove the annotation for site name and common name

api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedRowError.java
akashisama: Heh yeah weird state. Yeah,. add the field to DDL: no point having it lying around if it's going to be in an unusable state when we need it.

utas-tajona:
updated

utas-tajona:
added private accessor
acc

akashisama (approved):

PR #113, opened 2021-01-14
Feature/add eslint

utas-tajona:


ui/src/components/data-entities/customTooltip.js
```diff
@@ -17,19 +18,20 @@ export default class CustomTooltip extends Component {
...
-
-    return (<div className={"custom-tooltip ag-tooltip-hiding"} ></div>)
-
-
+    return (<div className={'custom-tooltip ag-tooltip-hiding'} ></div>);
   }
-}
\ No newline at end of file
+}
+
```
pmbohm, line 34: Whats this for?

ui/src/components/layout/AuthState.js
```diff
@@ -32,10 +32,10 @@ class AuthState extends React.Component {
                     variant="text"
                     color="secondary"
                     size="small"
-                    title={"Log out"}
+                    title={'Log out'}
                     style={ basicButton}
                     startIcon={<VerifiedUserIcon />}
-                    onClick={this.openLogout}> Logged in as '{ username }'</TopbarButton>
+                    onClick={this.openLogout}> { username }</TopbarButton>
```
pmbohm, line 38: Is this intentional?

ui/src/components/data-entities/customTooltip.js
utas-tajona: It would be good to know all the possibilities of  "value".

I left "any" because I saw the typedef which means it might comes as something else than 'object'

pmbohm: Yes you did start to explain

ui/src/components/layout/AuthState.js
utas-tajona: hmm it must come from a   different branch.  

It is good to mention 'Logged in as'  taking more space than anything else. showing the username with the icon 'check' is already a strong sign of being logged in.

utas-tajona: but the ' i will have to changed to &squo;

pmbohm: Ok i'll go with your change

pmbohm:
Overall a useful feature by the sounds of it

PR #116, opened 2021-01-14
Details view for stagedJobs

pmbohm:
Add non generic action link to GenericDetailsView


pmbohm:
For https://github.com/aodn/backlog/issues/2176

pmbohm:
This is the basics for the details view specifically for Jobs

ui/src/components/data-entities/GenericDetailsView.js
```diff
@@ -30,6 +30,13 @@ const renderError = (msgArray) => {
   return (msgArray.length > 0) ? <><Box><Alert severity="error" variant="filled">{msgArray}</Alert></Box></> : <></>;
 };
 
+const nonGenericEntities = {
+  'stagedJobs': {
+    linkLabel: 'Edit Job  ',
+    linkPath: 'PATHTOBECONFIRMED/stagedJobs/{}'
+  }
+};
+
```
utas-tajona, line 39: can you detail why stagedJob isn't a generic entity?

link label 'edit' should be enough

pmbohm: This demonstrates it will fit the edit endpoints. I doubt that will be a generic edit form

ccmoloney: There is no requirement to Edit the jobs.  They are created by the upload/stage process & then only the STATUS field is updated by the data ingest backend process.

PR #117, opened 2021-01-15
moved config file out of src

utas-tajona:


pmbohm:
Did you mean to move these files? Not delete them

utas-tajona:
it's a mistake the config file should be at the root of theUI project not in SRC

PR #118, opened 2021-01-18
Revert "moved config file out of src"

pmbohm:
Reverts aodn/nrmn-application#117

You dont merge your own pull requests here.

Additionally there is a previous pull request containing fixes created by the introduction of linting that will be ignored by separately deleting these files and re-creating elsewhere


utas-tajona:
it wasn't approved? how did it let me merged it ?? there was 3 green ticks

akashisama:
@utas-tajona sadly we can't force reviews for merges on repos with automatic version bumping due the way we implemented the version bumping (we need jenkins to be able to push to master :(). It's the sort of thing that github could easily fix at any moment, fingers crossed (they just need to add an "excluded users" option).

PR #119, opened 2021-01-18
js lint config

utas-tajona:
this should enable log/debugger in dev @pmbohm


utas-tajona:
if the build pass it should be ready

PR #122, opened 2021-01-27
error login and redirect without reloading the page

utas-tajona:


utas-tajona:
https://github.com/aodn/backlog/issues/2504

ccmoloney (approved):

PR #126, opened 2021-02-01
job-list v2, missing job-view

utas-tajona:


utas-tajona:
currently it's using the old job-view which make it functionall but I think it shouldn't use the generic view to display job logs

ui/src/components/job/jobMiddleware.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+function* getJobs() {
+    try {
+        const response = yield call(getEntity, 'stagedJobs');
+        const data = response.data?._embedded.stagedJobs || [];
+        if (data)
+            yield put(jobsFinished(data));
+        else {
+            yield put(jobsError([{ message: 'Service unavailbe.' }]));
```
nspool, line 16: typo in service unavailable

nspool (commented):
Looking at the description of issue https://github.com/aodn/backlog/issues/2603 while reviewing this PR:
- The Job Type column is missing, Should be INGEST or CORRECTION
- Also Does there need to be some sort of identifier for CORRECTION jobs?
- There are no filters implemented, such as for user
- The Build status 'FAILED' is aligned differently to 'STAGED'
image

On the Job page:
- There do not appear to be any log entries displaying
- Properties are showing with the camelcase variable name instead of a display name eg. "fullName" instead of "Full Name"
- Edit button throws the error: `SecurityError: Failed to execute 'pushState' on 'History': A history state object with URL 'http://validation/28' cannot be created in a document with origin 'http://localhost:3000' and URL 'http://localhost:3000/view/stagedJobs/28'.`

ccmoloney:
@utas-tajona Can you please address these issues as a matter of priority?

The 'identified for CORRECTION jobs' we can leave for now - we still need to decide on what the best identifier is for those jobs.  The user won't be able to create CORRECTION jobs yet anyway, so we have some time.

utas-tajona:
view page isn't implemented. it will be a seperated PR in order to keep the PR small

nspool:
@utas-tajona build appears to be failing.

nspool:
A few issues getting the filters to work:
- Initiator filter needs to be able to select a user
- Last Updated and Created filters should be able to specify a datetime range
- Filter "type" should be capitalised
- All the other filters can be hidden since they're not required

nspool:
Initiator filter is now working, but still not able to get the Last Updated or Created Date to filter. 
What I'm doing: I have a job with Last Updated as `28/01/2021, 08:52:46`. Going to Last Updated filter and select Greater Than 01/01/2021. Expecting that job to appear but instead the table returns zero results.

utas-tajona:
date getValue return a date now, which make the filter happening. noted  that we don't have time filter in ag-grid. so  04/02/2021 11:00 am is greater than 04/02/2021

nspool (approved):

PR #129, opened 2021-02-03
Build fails on react lint warnings

pmbohm:
The react warnings are still warnings in the dev environment while causing a build failure with `mvn clean package`


jonescc (approved):
LGTM

PR #130, opened 2021-02-03
Generic form changes

nspool:
- Hide ID when displaying location and diver entities
- Make new location active by default
- Remove the id from the generic form header


pmbohm (approved):
Killed it stone dead

PR #142, opened 2021-02-11
Store user roles from JWT, hide options if not logged in and add login button to TopBar

nspool:
Also removed the Settings menu and moved the theme setting to the side bar, removed the padding around the grids and fixed the dark-mode theme note applying to some of them, and also ran Prettier over the changed files.


utas-tajona:
I wasn't aware of this  github issue but we don't have to hide anything.

We need to implement a react-route PrivateRoute  that redirect user to /login if not login. we can stll reach 
/upload or /home if not loggin

PR #143, opened 2021-02-11
Custom template to implement ObservableItem form layout

nspool:
See `ObservableItemTemplate` for an example of a custom layout that still uses the components provided automatically by the JSON schema.


utas-tajona (approved):
LGTM

utas-tajona:
one last conflict and it's good to go

PR #131, opened 2021-02-04
Add custom list endpoint for site list view

jonescc:
spring data rest doesn't generate the correct metadata (json schema) for a spring data rest endpoint that uses excerpts or projections so just define it in a controller.

PR #144, opened 2021-02-11
Display required subset of observable item attributes in list view

jonescc:
Includes/requires  https://github.com/aodn/nrmn-application/pull/131


jonescc:
Yes, can't leave pull requests too long.   I'll fix

jonescc:
Conflicts resolved by revert

pmbohm (commented):
User see 'Clazz' not 'Class' in list header

jonescc:
> User see 'Clazz' not 'Class' in list header

I think that's an issue with the generic list code.  It doesn't look to be using the title defined in the json schema in preference to the attribute name.

jonescc:
I guess I can change the json attribute name which makes sense - it doesn't need to be restricted by java identifier naming - so will do that.

jonescc:
Fixed.

PR #147, opened 2021-02-12
Clean and divers fix

utas-tajona:
- adding reset state when page component died
-  minor fix on job entity
-   fix file with empty space
-  adding batch insert for saveAll()


api/src/main/resources/application.properties
```diff
@@ -50,3 +50,7 @@ app.api.version=@project.version@
 app.cors.max_age_secs=3600
 spring.servlet.multipart.max-file-size=600MB
 spring.servlet.multipart.max-request-size =600MB
+spring.jpa.properties.hibernate.generate_statistics=true
+spring.jpa.properties.hibernate.jdbc.batch_size=200
+spring.jpa.properties.hibernate.order_inserts=true
+hibernate.cache.use_query_cache=true
```
nspool, line 56: What are these changes related to?

ui/src/index.css
```diff
@@ -46,4 +46,12 @@ header a:hover {
...
     height: 18px;
     line-height: 18px;
+}
+
+.MuiBackdrop-root {
+    z-index: 200 !important;
+}
+
+.MuiDialog-paper {
+    z-index: 201 !important;;
```
nspool, line 56: Better not to use the `!important` directive or specify material ui css directly. Is this something that could be done in themes instead?

ui/src/components/store.js
```diff
@@ -19,9 +19,9 @@ import { jobReducer } from './job/jobReducer';
 import jobWatcher from './job/jobMiddleware';
 
 const initialiseSagaMiddleware = createSagaMiddleware();
+const isDev = (process.env.NODE_ENV == 'development');
 
-const middleware = [
-  ...getDefaultMiddleware(),
+const middleware = (isDev) ? [...getDefaultMiddleware(), initialiseSagaMiddleware] : [
```
nspool, line 24: What is this change for?

nspool (commented):
Looks good. Can you pelase check that the Prettier style is being applied to the UI JS files?

api/src/main/resources/application.properties
utas-tajona: yes this will increase speed at upload and not failed on the edge env

ui/src/index.css
utas-tajona: it's litteray the same rule usine z-index being override so it's above ag-grid

ui/src/components/store.js
utas-tajona: getDefaultMiddleware check lots of dev stuff like is the state immutable.
it slow down considerably the UI. 

Good for dev bad for the rest

api/src/main/resources/application.properties
```diff
@@ -50,3 +50,7 @@ app.api.version=@project.version@
 app.cors.max_age_secs=3600
 spring.servlet.multipart.max-file-size=600MB
 spring.servlet.multipart.max-request-size =600MB
+spring.jpa.properties.hibernate.generate_statistics=true
+spring.jpa.properties.hibernate.jdbc.batch_size=200
```
jonescc, line 54: You could increase this - 200 is actually a low number - we use 1000 for our harvesters

jonescc (commented):
Backend changes LGTM.   @nspool?

nspool:
Looks good, but we should try and avoid using `!important` in future CSS. Can you please check that the Prettier style is being applied to the UI JS files? It should be using the prettierrc in the UI directory.

nspool (commented):
Just needs formatting and merge conflicts resolved and should be good to merge.

nspool (approved):

PR #148, opened 2021-02-12
delete job

utas-tajona:
backend & frontend

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedJobLog.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedJobLog.java
@@ -10,19 +10,7 @@ import lombok.NoArgsConstructor;
 import lombok.Setter;
 import org.hibernate.annotations.CreationTimestamp;
 
-import javax.persistence.Column;
-import javax.persistence.Entity;
-import javax.persistence.EnumType;
-import javax.persistence.Enumerated;
-import javax.persistence.FetchType;
-import javax.persistence.ForeignKey;
-import javax.persistence.GeneratedValue;
-import javax.persistence.GenerationType;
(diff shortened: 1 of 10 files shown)


utas-tajona:
https://github.com/aodn/backlog/issues/2640

api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedJobLog.java
```diff
@@ -37,7 +25,7 @@
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "staged_job_log_id_seq")
     private Long id;
 
-    @ManyToOne
+    @ManyToOne(cascade = CascadeType.ALL)
```
jonescc, line 28: Did you need this? I don't think its required. Normally cascade is used on the parent only.

api/src/main/java/au/org/aodn/nrmn/restapi/model/db/StagedRow.java
```diff
@@ -111,7 +112,7 @@
     private Map<Integer, String> measureJson;
 
     @JsonIgnore
-    @ManyToOne(fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.LAZY, cascade = CascadeType.ALL)
```
jonescc, line 115: And here.

jonescc (commented):
LGTM.  @nspool ?

nspool (commented):
The JS style seems to differ from the .prettierrc config in the UI project. Can you please check that it is using the right config?

utas-tajona:
i've setup prettier on vscode and run it on the js file of this PR

nspool (commented):
Looks like it's matching the config.

PR #155, opened 2021-02-16
Fix duplicates being returned when validating jobs

jonescc:
Validation process was looking up associated rows by reference (file name) so all rows uploaded for that reference (file name) were being included in validation including ones from other jobs.

Also fixed some issues with Integration tests.


utas-tajona (approved):
LGTM

PR #156, opened 2021-02-17
Cache findbycriteria responses made in a request

jonescc:
As per https://okraskat.github.io/caching-per-request-with-spring/

Improves validation performance  for https://github.com/aodn/backlog/issues/2661

Added for Diver and Site only - not expecting any improvement for species.


utas-tajona (commented):
how long the value get cached ?
in a near future we might need to clear cache by entity (if someone add a diver all cached diver should be deleted)

jonescc:
> how long the value get cached ?

The cache is a request cache so values are cached for the request only.

utas-tajona (approved):

PR #157, opened 2021-02-17
Use hibernate L2 cache and query cache instead of custom caching solution

jonescc:
Hibernate manages the cache when updates are made and so this will speed up species checks as well over time.  

Currently using ehcache memory cache but could use more persistent shared cache in future.

Refer https://www.baeldung.com/hibernate-second-level-cache


utas-tajona (approved):
LGTM

PR #160, opened 2021-02-19
Remove dark mode

nspool:
- Move the remaining setting to the side menu
- Modify AuthState to redirect to login if logged out 
- Run Prettier on changed files
- Removed files  `9  .eslintignore`, and spurious `yarn.lock` and `package.json` files that are not needed


jonescc (commented):
Tested OK for me.

utas-tajona (approved):

PR #173, opened 2021-02-26
adding delete rows API

utas-tajona:
adding the UI link soon


jonescc (commented):
Just testing this and when I delete multiple rows using right click - delete selected row.  The rows appear to be deleted but when I validate most of them come back.  Looking at the requests sent to the backend - the same id is sent each time.

Also, this doesn't appear to be a 'Saveable' operation - the save button is not required to send delete operations to the backend.   Seems inconsistent to batch up some changes to save and not others.

ccmoloney:
@utas-tajona Are you still working on this - or is it ready for re-review?

nspool (commented):
Rows are deleting immediately from the DB. Should it wait until Save is pressed? Also Undo is not working for delete.

ccmoloney:
I think we want deletes to only happen in the front-end until Save is invoked by the user.  Deleting rows should trigger the Save button to become active

ccmoloney (approved):

PR #174, opened 2021-03-01
Fix issue running integration tests in github actions

jonescc:
- Upgrade to a recent version of testcontainers containing a fix for a recent issue pulling test containers from docker repository.
- Trap and report errors thrown when starting up the postgres test container.  We want tests to fail if the container doesn't startup instead of silently skipping them.
- Fix broken tests not picked up because of these issues.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -14,6 +14,7 @@
 
     <properties>
         <java.version>1.8</java.version>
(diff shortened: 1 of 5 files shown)


nspool (approved):
Looks good and happy to merge.

jonescc:
Good enough for me ;-)

PR #177, opened 2021-03-02
Add endpoints for ui_protection_status and ui_mpa

jonescc:
Also remove template_code from ObservableItem - its been removed from the db (merged with letter_code)

image

image


nspool (approved):
Tested new endpoints against my PR and the correct values are being returned.

PR #179, opened 2021-03-03
limit redux interaction with ag grid

utas-tajona:


utas-tajona:
http://nrmn-ag-grid-api-reduxless.dev.aodn.org.au/validation/1

jonescc:
Just shows "No data" for me

jonescc:
Not redirecting to login - OK when I login.

jonescc:
@bpasquer do you want to have a look at this as well?

jonescc:
For https://github.com/aodn/backlog/issues/2707 - missing undo of cut.

jonescc (commented):
Some issues during testing

- Undo stops working after clicking on an error and then clearing the error filter or performing any cut operation
- Going to the validate endpoint when not logged in does not redirect to login form 
- Save button is enabled when viewing the file before making any changes

jonescc:
Build failing due to console.log's

jonescc:
Also can you rebase against master as this will fix some of those validation errors.

nspool (changes requested):
Just testing locally, noticed that validation errors are not being highlighted unless the column is refreshed. Also, clicking on the validation error isn't highlightly it in the grid.

utas-tajona:
Validation error in the grid shouldn't be highlighted , there is no spec about that right now.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

ccmoloney:
@utas-tajona 
> Validation error in the grid shouldn't be highlighted , there is no spec about that right now. Get Outlook for 

@bpasquer Do you have a view on this?  I think that the highlighting was probably quite useful as a visual indicator - allows you to scan the data and see what's not passing validation.  Was there a technical reason that we removed it?

nspool:
There seems to be an issue where there column is not immediately refreshed if it has an error. See column containing "Nadia" that could not be found. Error is not highlighted until after scrolling off and back again:
nadia could not be found

utas-tajona:
I think validation need much more work. Once we get more validators implemented few questions will come: 
- the way group validation and row validation will work.
-  Does entities error get link to add new reference data?
-  what level of aggregation do we need? 
- How do we group the different validation made?

This need to be discuss with Lizz and Toni first

This PR is not related to validation but more about the Ctrl c/x/p/z/y and the speed of editing.

utas-tajona:
http://nrmn-ag-grid-api-reduxless.dev.aodn.org.au/validation/2

utas-tajona:
@ccmoloney There was never any highlight before.
the previous behaviour that i implemented was to filter the row include for each validation on click.

the problem is Ag-Grid doesn't like fitering via custom filter and break ctrl-z.  so instead of filtering, it's scrolling to the first row concern and "select" all rows concerned.

None of these behavior have been discuss or specify.   I would suggest to discuss with liz and Toni first.

utas-tajona:
@nspool  I have force the all grid to reset instead of trying to refresh all cells. it works but slower.

ccmoloney:
When i say "highlighting" i meant the text colour for cells that aren't passing validation.  there is definitely some sort of refresh issue when the validate function returns - it's not changing the text colour for those cells until you scroll away and then back

utas-tajona:
it should be updated on the  branch instance

nspool (approved):
Now works for me.

PR #185, opened 2021-03-10
Implement Site Clone

nspool:
Also: 
- Adds tool tips to the action buttons on the table
- Fixes a bug where returning to the Site table using a back arrow link show the table with only the Edit button visible
- Updates UI unit test to use the new entity template format


nspool:
Need to fix tests before review.

utas-tajona (approved):
LGTM

PR #186, opened 2021-03-10
SQL script to insert a data and power user

nspool:
Handy SQL script for local development. Adds a power user with email `power@example.com` and data officer with email `data@example.com`. Password  for both users is `password`.


jonescc (commented):
This will be included in the production war which seems unnecessary (anything in main/resources is included).  I'd suggest putting it in api/scripts or something similar

nspool:
@jonescc moved to api/scripts and renamed to `add_dev_users.sql`.

jonescc (approved):

PR #187, opened 2021-03-11
lowercase on groupby

utas-tajona:


jonescc:
So do we still want to merge this or are we going to convert everything to lowercase when staging?  TBH - if case doesn't matter then keeping it consistently lower or uppercase in the system makes perfect sense to me.   @ccmoloney @bpasquer ?

bpasquer:
We should keep the case insensitive lookups but for consistency SiteCodes,  Divers buddy and PQs initials and even Direction should be ingested in the DB as uppercase.

jonescc:
> We should keep the case insensitive lookups but for consistency SiteCodes, Divers buddy and PQs initials and even Direction should be ingested in the DB as uppercase.

Probably should be to upper case anyway as that what  al the reference data is.   

Lets deal with it in a separate issue.

I'm happy to merge this one.

jonescc:
Actually @utas-tajona - sounds like group by should be upper case.

utas-tajona:
for the group by it shouldn't  matter

jonescc (approved):

PR #188, opened 2021-03-11
Fixes aodn/backlog#2806

jonescc:
Remove blacklisted token when logging in if the same token is generated again.

Use a separate user for authorisation controller tests - prevents issues using that user again within the same second in subsequent tests - see aodn/backlog#2806.

In future we could actually login instead of just generating a token to better test actual authentication and authorisation during tests.


jonescc:
Should also remove blacklisted token on login for scripted interactions.

nspool (approved):

PR #191, opened 2021-03-12
Ensure new sites are inactive on create

nspool:


nspool:
Updated to make the column not null and false by default.

api/src/main/resources/sql/migration.sql
```diff
@@ -551,7 +551,7 @@ CREATE TABLE nrmn.site_ref (
     wave_exposure integer,
     slope integer,
     site_attribute jsonb,
-    is_active boolean
+    is_active boolean NOT NULL
```
jonescc, line 554: This also needs to be done in https://github.com/aodn/NRMN  - contains the master ddl for core tables.  But can address that separately.

jonescc (approved):

api/src/main/resources/sql/migration.sql
jonescc: https://github.com/aodn/NRMN/blob/master/db-migrate/ddl/CreateTables.sql#L42

bpasquer: > This also needs to be done in https://github.com/aodn/NRMN - contains the master ddl for core tables. But can address that separately.

Yes I've prepared a PR  to set all `is_active` attributes NOT NULL in the migration code

PR #193, opened 2021-03-15
Minor UI Fixes

nspool:
- Location Edit and Diver Edit not redirecting correctly if there is no error on Save
- Location Add: Is Active is showing as False when this value defaults to True


jonescc (approved):
LGTM

PR #197, opened 2021-03-16
Misc UI fixes

nspool:
- Fix MPA and Protection Status fields not updating when using values not in the drop-down
- Relief, Slope, Wave Exposure and Currents fields: change values to  [1, 4] instead of [0, 3], matching database values and make nullable
- Fix React uncontrolled component warnings
- Rename entity classes to be more consistent
- Minor refactoring and removal of unused code


jonescc (approved):
LGTM

PR #200, opened 2021-03-17
Observable Items - List

nspool:
- Adds a ObservableItemListItemRepository to return list entries from `ep_observable_items` instead of `observable_item_ref`
- Wires up UI and implements an optional `sort` property to disable sorting on all columns except name and type.

http://nrmn-obs-item-list.dev.aodn.org.au/


jonescc (approved):
LGTM

PR #201, opened 2021-03-17
Species lookup - WORMS and Observable item fuzzy search endpoint

jonescc:
For aodn/backlog#2686


jonescc:
Still need to remove extraneous text as per backlog item before performing searches

jonescc:
Should have mentioned that this PR requires postgres extension pg_trgm to be installed for the similarity matching to work - I've added that to the postgrescontainer setup for tests but if you want to try this locally you'll need to install that.

Requires https://github.com/aodn/chef-private/pull/3775 which installs it in nrmn_dev/edge to deploy there.

nspool (approved):

PR #204, opened 2021-03-24
add pq_diver_id to survey_aud table

bpasquer:


jonescc (commented):
Looks good - just need to add the survey changes to migration.sql when  the migration is done.  

I'll add that

jonescc (approved):

PR #205, opened 2021-03-24
fixing loading/submit enable

utas-tajona:


jonescc (commented):
No longer shows loading icon, but when I fix my last validation error the validation error is not cleared and if I refresh and validate it returns no errors but the submit button isn't enabled.

jonescc:
Couldn't build

jonescc (approved):
Submit button is now correctly enable on successful validation but submitting doesn't do anything.

Still this is better than what is currently happening

jonescc:
Looks like its posting to http://localhost:8080/api/ingestion/ingest/2 - and this is not allowed - get a 405

jonescc:
Also, if you make changes after validating - the submit button is not disable/save enabled.

PR #212, opened 2021-03-29
Indicate if species superseded in species search

nspool:
Superseded species names appear slightly greyer to other search results and clicking them will not autofill the form but display a warning containing the superseded name. 

Switched the order of the NRMN and WoRMS search so that WoRMS is now the default.

Also made the search component much wider and taller to avoid scrolling.


jonescc (approved):
LGTM

PR #213, opened 2021-03-29
Observable Item fixes and improvements

nspool:
- Rename 'Superseded By' fields to 'Superseded'
- Restore class column in table view
- Improve ObservableItemGetDto mapping

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/dto/observableitem/ObservableItemGetDto.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/dto/observableitem/ObservableItemGetDto.java
@@ -38,10 +38,10 @@ public class ObservableItemGetDto {
     private String supersededBy;
 
     @Schema(title = "Superseded Names")
-    private String supersededByNames;
+    private String supersededNames;
 
     @Schema(title = "Superseded IDs")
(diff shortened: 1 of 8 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/dto/observableitem/ObservableItemGetDto.java
```diff
@@ -37,11 +37,11 @@
...
     private String supersededBy;
 
-    @Schema(title = "Superseded By Names")
+    @Schema(title = "Superseded Names")
     private String supersededByNames;
 
-    @Schema(title = "Superseded By IDs")
-    private String supersededByIDs;
+    @Schema(title = "Superseded IDs")
+    private String supersededByIds;
```
jonescc, line 44: We should change the variable references as well - they're supersededIds and names not supersededByIds and NAmes

jonescc: My mistake by the way.

nspool: Agree.

PR #218, opened 2021-04-07
General UI Improvements

nspool:
Enable column autosizing & remove column menu
Change Jobs List filters to match other screens
Display 'Other Attributes' in a table instead of a string
Increase form width and general layout changes to improve use of space.
Change link on Job Upload success to redirect to Job List

Also moved all of the templates into a `templates` directory and remove unused dependency `lodash`.

Stack available at http://nrmn-ui-tidy.dev.aodn.org.au/home


utas-tajona (approved):
LGTM

PR #227, opened 2021-04-15
logout if timeout expire 24h

utas-tajona:


jonescc:
Better - redirects to login page when token expires.   But does so while you are actively using the app.   I think animal tracking may refresh the token so that if you are actively using the app you don't get logged out.   But that can be the subject of another backlog item @ccmoloney ?

jonescc:
Actually couldn't find any animal tracking pr like that.  Its a nice to have, will merge this when build reruns successfully.

jonescc (approved):

PR #228, opened 2021-04-16
aodn/backlog#2730 display duplicate survey details

jonescc:
Display each duplicate survey in errors

image


utas-tajona (commented):
LGTM

PR #230, opened 2021-04-16
Survey API: Implement filters needed for corrections page

nspool:
The ugly casting appears to be necessary to work around null values being converted to type bytea. Please let me know if there is a better way of doing this.


utas-tajona (changes requested):
looks good it just needs a test on the filter

nspool:
@utas-tajona added tests however they are failing due to the fact that materialized view `ep_site_list` isn't populated. Leaving in but commented out for now.

ccmoloney:
@bpasquer any thoughts on why ep_site_list wouldn't be populated?  @nspool you mean in the dev database?

nspool:
@ccmoloney yes, in the db used for running the unit tests.

bpasquer:
No no idea why it is not populated. Everything looks ok is nrmn_migration.

ccmoloney:
We think it's something to do with how CJ has set up the unit tests to run.  It seems fine in nrmn_dev & nrmn_edge too.

utas-tajona (commented):
LGTM

PR #236, opened 2021-04-27
no species found

utas-tajona:


utas-tajona:
https://github.com/aodn/backlog/issues/2933

api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/RawValidation.java
```diff
@@ -91,6 +91,7 @@
                         Tuple2.of("Latitude", new DoubleFormatValidation(StagedRow::getLatitude, "Latitude")),
                         Tuple2.of("Longitude", new DoubleFormatValidation(StagedRow::getLongitude, "Longitude")),
                         Tuple2.of("Direction", new DirectionDataCheck()),
+                        Tuple2.of("found",new SpeciesNotFoundCheck()),
```
jonescc, line 94: Need an import for this - getting a build error.

jonescc (changes requested):
Looks good to me - but we need to fix that build error.

jonescc (approved):
Works OK for non-zero measures for species not found but doesn't check total/inverts as per spec.

Merging as the measure checks work and noting this on backlog item for @bpasquer to clarify.

jonescc:
Also I note that we get more than one species heading now - one for this check and one for Species not found checks.  Not sure if that's important.

jonescc:
image

PR #242, opened 2021-04-28
Job upload: Rewrite sheet parsing to use SAX parser

nspool:
Rewrite job upload to use a one-pass SAX parser instead of validating headers seperately
Enable Hibernate batched inserts and update schema to dramatically increase insert speed
Fixes error messages to advise re-saving unsupported files as Excel Workbook (*.xls)
Remove some unused models

Also imports an extra row from the file "Copy of Rottnest 2019_compiled data.xlsx" that is currently being skipped.

Stack available using a new database restored from yesterday's edge backup: http://nrmn-upload-improvements.dev.aodn.org.au/jobs

Also against nrmn_dev database: http://nrmn-upload-improvements-devdb.dev.aodn.org.au/jobs

nrmn_dev has the necessary schema changes made.


api/src/main/java/au/org/aodn/nrmn/restapi/service/SpreadSheetService.java
```diff
@@ -42,150 +38,47 @@
 
     @Autowired
     private S3IO s3client;
+    
+    public Validated<ErrorInput, List<StagedRow>> stageXlsxFile(MultipartFile file, Boolean withInvertSize) {
+        try {
+            InputStream inputStream = file.getInputStream();
```
jonescc, line 44: This needs to be closed in a finally to ensure its always closed otherwise we can un out of file system handles (or whatever they're called)

jonescc: Better yet use a try with resources block

jonescc: try (InputStream inputStream = file.getInputStream()) {
    ...
}

api/src/main/java/au/org/aodn/nrmn/restapi/service/SpreadSheetService.java
```diff
@@ -42,150 +38,47 @@
...
+            OPCPackage opcPackage = OPCPackage.open(inputStream);
+            XSSFReader xssfReader = new XSSFReader(opcPackage);
+            ZipSecureFile.setMinInflateRatio(0.0d);
 
+            SurveyContentsHandler surveyContentsHandler = new SurveyContentsHandler((withInvertSize) ? longHeadersRef : shortHeadersRef);
```
jonescc, line 49: I know you didn't name it this but the option shouldn't be called withInvertSize - it wqould be better called withExtendedSizes - it applies to both fish and invert sizing and refers to extended sizes being used for both.  Its been a source of confusion for me up until recently until clarified by @bpasquer.  (Thanks Bene!)

jonescc: But you don't have to change this now.  I think its referred to as that in the api and must be referenced in the UI as well so would require a few changes.  Just ranting

nspool: Will change the variable name in this PR but change the parameter name with the upload page PR.

api/src/main/java/au/org/aodn/nrmn/restapi/service/SpreadSheetService.java
```diff
@@ -42,150 +38,47 @@
...
-                    stagedRow.setPos(index - 1);
-                    if (dataSheet.getHeader().size() == longHeadersRef.size()) {
-                        stagedRow.setM2InvertSizingSpecies(_getCellValue(row.getCell(headerMap.get("M2 Invert Sizing Species")), eval, fmt));
-                        stagedRow.setL5(_getCellValue(row.getCell(headerMap.get("L5")), eval, fmt));
-                        stagedRow.setL95(_getCellValue(row.getCell(headerMap.get("L95")), eval, fmt));
-                        stagedRow.setIsInvertSizing(_getCellValue(row.getCell(headerMap.get("Use InvertSizing")), eval, fmt));
-                        stagedRow.setLMax(_getCellValue(row.getCell(headerMap.get("Lmax")), eval, fmt));
+            Iterator<InputStream> sheets = xssfReader.getSheetsData();
+            while ((sheets instanceof SheetIterator) && sheets.hasNext()) {
+                InputStream i = sheets.next();
```
jonescc, line 62: This is OK, but would be more consistent with other usages if it was put into a try with resources block

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyContentsHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        measureJson.clear();
+    }
+
+    @Override
+    public void endSheet() {
+        if (stagedRows.size() > 0)
+            result = Validated.valid(stagedRows);
+        else
+            result = result != null ? result : Validated.invalid(new ErrorInput("Empty DATA sheet", "sheet"));
+        SheetContentsHandler.super.endSheet();
```
jonescc, line 69: Is this right - I would have expected invoking the inherited endSheet on this instance i.e. super.endSheet() 
This is a static reference and so wouldn't have any  state if its required.

jonescc: May be OK but just super.endSheet() is what I'm used to

nspool: This was boilerplate added by the editor. POI code samples aren't calling this so I'll remove it.

nspool:
Added additional check to fail when additional unexpected headers appear in file, such as when an extended sheet is ingested with extended sizing checkbox set. Also added unit test.

jonescc (approved):
Looks good - fixes a few issues and super fast.

PR #244, opened 2021-04-29
adding structure to support formatted global check

utas-tajona:


jonescc:
Looks good.  Just about to test but you'll need to fix the conflicts introduced when I merged your other PR.

api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/global/formatted/Method3QuadratsSum.java
```diff
@@ -0,0 +1,44 @@
...
+import au.org.aodn.nrmn.restapi.model.db.StagedRowError;
+import au.org.aodn.nrmn.restapi.validation.StagedRowFormatted;
+import au.org.aodn.nrmn.restapi.validation.validators.base.BaseGlobalFormattedValidator;
+import cyclops.control.Validated;
+import lombok.val;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class Method3QuadratsSum extends BaseGlobalFormattedValidator {
```
jonescc, line 12: Can't see where this s instantiated - do we need to add it to the global format checks for RLS/ATRC or common

jonescc (commented):
Looks good.  Couldn't test the Method3Quadrat check as it doesn't look to be configured.  But if you fix the conflicts and the check is to be added later - I can merge

jonescc (changes requested):
Looks like its not building now. But conflicts have been resolved!

PR #260, opened 2021-05-04
Job Upload: Store an empty string or measurement instead of null

nspool:
Stack available at http://nrmn-ingest-remove-null.dev.aodn.org.au/


nspool:
Converting to draft PR to fix tests.

nspool:
Empty strings were being added for each of the measurement bins. Fixed so that only non-measurement columns are pre-populated with an empty string.

jonescc (approved):

PR #262, opened 2021-05-04
Job upload: Custom dataformatter to prevent ambiguously formatted dates

nspool:
Implements a DataFormatter to override US date formatting on ingest.

 Stack available at: http://nrmn-fix-ingest-date-formatting.dev.aodn.org.au/


api/src/main/java/au/org/aodn/nrmn/restapi/service/SpreadSheetService.java
```diff
@@ -40,11 +40,13 @@
     private S3IO s3client;
 
     public Validated<ErrorInput, List<StagedRow>> stageXlsxFile(MultipartFile file, Boolean withExtendedSizes) {
+
+        ZipSecureFile.setMinInflateRatio(0.0d);
```
nspool, line 44: Moved from below as this needs to be set *before* the stream is opened.

jonescc (approved):

PR #264, opened 2021-05-04
Fix save measures json and scroll on validation error

utas-tajona:


utas-tajona:
https://github.com/aodn/backlog/issues/2966

utas-tajona:
https://github.com/aodn/backlog/issues/2963

nspool (approved):

PR #273, opened 2021-05-07
rule + test

utas-tajona:


jonescc (approved):
LGTM - but it looks like that attribute probably needs to be editable now.  I'll raise a separate issue

PR #277, opened 2021-05-11
Observable Item Edit: Show only active species in `superseded by` dropdown

nspool:
Also adds includeSuperseded param to the api/search endpoint.

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/controller/SpeciesController.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/controller/SpeciesController.java
@@ -34,6 +34,7 @@ public class SpeciesController {
 
     @GetMapping
     public List<SpeciesDto> findSpecies(@RequestParam("species") String speciesName,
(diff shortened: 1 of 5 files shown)


nspool:
Tests have failed. Converting to draft.

jonescc:
We probably need to ensure that the superseded by name used exists as well - I can add any random text.   @ccmoloney?

nspool:
Updates tests and is ready for review again.

bpasquer:
> We probably need to ensure that the superseded by name used exists as well - I can add any random text. @ccmoloney?

The superseded_by fields should come from the database. So by default we can assume it exists.

jonescc:
> The superseded_by fields should come from the database. So by default we can assume it exists.

There's nothing stopping someone from accidently modifying the superseded by name entered and saving it or just typing in something they think is valid and saving it - there are no foreign key constraints in the db or any checks in the app.   So no we can't assume that its valid.

But that's a separate issue to what this pr was addressing.

PR #287, opened 2021-05-13
Site actif first ingest

utas-tajona:


api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
```diff
@@ -40,32 +40,51 @@
...
-                .depth(stagedRow.getDepth())
-                .surveyNum(stagedRow.getSurveyNum().orElse(null))
-                .direction(stagedRow.getDirection().toString())
-                .site(stagedRow.getSite())
-                .surveyDate(Date.valueOf(stagedRow.getDate()))
-                .surveyTime(Time.valueOf(stagedRow.getTime().orElse(null)))
-                .visibility(stagedRow.getVis().orElse(null))
-                .program(stagedRow.getRef().getStagedJob().getProgram())
-                .build()));
+        val siteSurveys = surveyRepository.findAll(
```
jonescc, line 63: Isn't enough just to set the sitecode to active if its not active - its being used on this survey - not sure why we are checking for existing survey's

api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/formatted/SpeciesBelongToMethodCheck.java
```diff
@@ -11,15 +11,15 @@
...
+        super("Species");
     }
 
     @Override
     public Validated<StagedRowError, String> valid(StagedRowFormatted target) {
 
         if (target.getSpecies()
                 .getObsItemAttribute()
-                .containsKey("is_M" + target.getMethod())) {
+                .containsKey("is_M" + target.getMeth                              od())) {
```
jonescc, line 22: ?

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
utas-tajona: a site could be set to false trough  the data reference. not  only when it get created. 
So I had to check if was never used before

api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/formatted/SpeciesBelongToMethodCheck.java
utas-tajona: updated

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
jonescc: Its not possible to set the site to be inactive or active  when adding or editing a site.   It gets defaulted to inactive when its created.

jonescc: I believe the intention of the isActive flag is to reflect whether there are any survey's associated with the site.  So the siteSurveys check below doesn't make any sense - we always want to set the site to be active if its inactive when creating a survey for that site.

@ccmoloney / @bpasquer can you clarify

utas-tajona: if it is then i will just check if it's false then

ccmoloney: I think the logic should be:

if isActive flag is FALSE
    If this is the first ingest for this site
        Set isActive flag to TRUE

jonescc: Doesn't isActive = false mean it is the first ingest for the site?

jonescc: So that flag is untrustworthy anyway.   Not sure why we're bothering.

ccmoloney: Only for new sites.  They may be able to deactivate old sites in the future

PR #293, opened 2021-05-17
empty measure should be valid

utas-tajona:


api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/formatted/SpeciesBelongToMethodCheck.java
```diff
@@ -9,16 +10,20 @@
...
         super("Species");
     }
 
     @Override
     public Validated<StagedRowError, String> valid(StagedRowFormatted target) {
-
-        if (target.getSpecies()
-                .getObsItemAttribute().entrySet()
+        val species =  target.getSpecies();
+        val attributes =  Optional.of(species).map(ObservableItem::getObsItemAttribute).orElseGet(Collections::emptyMap);
```
jonescc, line 25: ?  Not sure what this is trying to accomplish - Optional.of will throw a null pointer exception if species is null - but species can't be null here thanks to you modifications in a previous PR.    The null pointer exception was being thrown in systest because getObsItemAttribute was null which this doesn't address.

PR #311, opened 2021-05-19
Fix species epithet in WoRMS search results

nspool:
Remove the genus from the species to autofill the species epithet for new observable items.


utas-tajona (approved):
LGTM

PR #326, opened 2021-05-25
Fix 'not a Object[] or List\!' issue

jonescc:
Spring JPA find by example can't handle site's with geometries.  Use a custom method to simplify the required processing.


utas-tajona (approved):
lgtm

PR #328, opened 2021-05-25
Paginate species search results

nspool:
Also changes the NRMN search to do a substring match on species name instead of just checking the start of the string.

Stack at: http://nrmn-pagination.dev.aodn.org.au


jonescc:
Still not finding Ranella australasia for some reason.

jonescc:
> Still not finding Ranella australasia for some reason.

Only showing 2 pages but there are at least 3

jonescc (changes requested):
Looks like the api can return more pages than shown so looks like an issue in the front-end?

nspool:
Caused by the worm search method removing deleted records from the search results. Front-end interprets less than 50 results as being the final page. Removed filter so that all search results are returned.

jonescc (approved):

PR #333, opened 2021-05-26
aodn/backlog#3042  - Speed up ingestion

jonescc:
Ingest by survey/survey method so that we only have to look up survey/survey method once
And don't use find by example for looking up measures as this doesn't use the cache.  FindById makes more sense anyway.

Need to revisit ingestion tests as they don't make sense now after refactoring.


ccmoloney (approved):

jonescc:
ready for merge.  I'll update tests in a separate PR.

PR #343, opened 2021-05-28
Additional size class mappings

nspool:
http://nrmn-size-classes.dev.aodn.org.au/ using nrmn_dev db


jonescc (approved):
Will probably need to rejig this to use in validations as well.   But works as expected when ingesting.   I'll make sure the "BTW method 2 debris should be ingested as method 12" issue gets raised separately.

jonescc:
Also need to revisit species attributes warnings to take this into account - I think @bpasquer is raising another issue

bpasquer:
Yes @jonescc, see  https://github.com/aodn/nrmn-application/issues/345

PR #346, opened 2021-05-28
Ingest: Save debris as M12

nspool:
Stack deployed to http://nrmn-debris-m12.dev.aodn.org.au


jonescc (changes requested):
No longer creating any observations for debris.  May be a previous PR that did this?

jonescc:
Sounds like that's related to the invert size observations being dropped - I'll include it in the PR for that.

PR #354, opened 2021-06-01
Fix measure size validation

nspool:
Convert survey column to the corresponding size class before comparing
with the L5 and L95 bounds.

Add sizing labels for invert validation messages.


nspool:
Stack available at http://nrmn-measure-validation.dev.aodn.org.au

jonescc:
Actually that looks like the lmax calcs. Which wasn't a subject of this PR.  We'll need to fix that as well.   The others are likey because the IsInvertSizing isn't filled out correctly.  I'll check with Bene's test sheet.

jonescc:
l5/l95 messages look good.

PR #360, opened 2021-06-02
Validate 'use invert sizing' as Yes or No

nspool:
The 'OptionalBooleanValidation' is no longer needed and is removed. Tests also updated.


api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/RawValidation.java
```diff
@@ -40,169 +40,157 @@
...
+                rowFormatted.setCode(code);
+                rowFormatted.setDirection(direction);
+                rowFormatted.setTotal(total);
+                rowFormatted.setInverts(inverts.orElse(0));
+                rowFormatted.setMeasureJson(measureJson);
+                rowFormatted.setRef(ref);
+                rowFormatted.setSpeciesAttributesOpt(mayBeSpeciesAttributes);
+
+                if (isExtendedSizing) {
+                        val isInvertSizing = (Optional<Boolean>) values.get("IsInvertSizing").orElse(Optional.empty());
```
jonescc, line 178: ??  I thought we weren't using optionals any more  - doesn't this throw a class cast exception?

jonescc: And this value is actually optional when isextendedsizing is false

nspool: Changed the validation step but not the subsequent uses. Converting back to draft.

PR #362, opened 2021-06-04
Improve error display on ingest submit

nspool:
Display any ingest errors as an alert on the datasheet page.


ccmoloney (approved):

jonescc (approved):
Look good

image

PR #363, opened 2021-06-04
Unsized species validation

nspool:
Do not validated species sizing when L5-L95 and Lmax are zero or null

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/row/formatted/MeasureBetweenL5l95.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/row/formatted/MeasureBetweenL5l95.java
@@ -33,10 +33,10 @@ public class MeasureBetweenL5l95 extends BaseFormattedValidator {
         }
 
         val speciesAttributes = target.getSpeciesAttributesOpt().get();
-        val l5 = speciesAttributes.getL5();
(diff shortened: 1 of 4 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/row/formatted/MeasureBetweenL5l95.java
```diff
@@ -36,8 +36,8 @@ public MeasureBetweenL5l95() {
         val l5 = speciesAttributes.getL5();
         val l95 = speciesAttributes.getL95();
         val measureJson = target.getMeasureJson();
-        if (measureJson.isEmpty() || l5 == null || l95 == null)
-            return Validated.valid("No data");
+        if (measureJson.isEmpty() || (l5 == null && l95 == null) || (l5 == 0 && l95 == 0))
```
jonescc, line 39: Can result in null pointer exceptions?

```
select count(*) from nrmn.ui_species_attributes
 where (l95 is null and l5 is not null) or (l5 is null and l95 is not null)

34
```

nspool: Will change so that if any value is null it is treated as zero, and ignore zero values when evaluating validation rules.

PR #366, opened 2021-06-06
Skip size validation on M3, M4 and M5 species

nspool:


api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/row/formatted/MeasureBetweenL5l95.java
```diff
@@ -28,11 +28,7 @@ public MeasureBetweenL5l95() {
         if (skipMethods.contains(target.getMethod())) {
             return Validated.valid("M3, M4, M5 species");
         }
-
-        if(!target.getRef().getStagedJob().getIsExtendedSize()) {
```
jonescc: Need to modify code below to handle target.getIsInvertSized now possibly being null.

jonescc: Or better ye - only refering to it if extendedSizing applies

api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/row/formatted/MeasureBetweenL5l95.java
```diff
@@ -40,7 +40,7 @@ public MeasureBetweenL5l95() {
         if (measureJson.isEmpty() || (l5 == 0 && l95 == 0))
             return Validated.valid("No expected sizing");
 
-        boolean isInvertSized = target.getIsInvertSizing();
+        boolean isInvertSized = target.getIsInvertSizing() ? target.getIsInvertSizing() : false;
```
jonescc, line 43: I'm pretty sure this will throw a null pointer exception if getIsInvertSizing is null

jonescc: Actually throws this:

    incompatible types: <null> cannot be converted to boolean

nspool: Will rework this check to be more robust.

PR #372, opened 2021-06-07
Block value can only be 0,1 or 2.

nspool:


api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/RawValidation.java
```diff
@@ -71,7 +71,7 @@
                                                                 ValidationLevel.BLOCKING)),
                                 Tuple2.of("Block",
                                                 new IntegerFormatValidation(StagedRow::getBlock, "Block",
-                                                                Arrays.asList(0, 1, 2, 10))),
+                                                                Arrays.asList(0, 1, 2))),
```
jonescc, line 74: This is for ATRC and RLS - does the same rule apply to both?

nspool: Will check with @bpasquer

bpasquer: Yes the rule applies to both

PR #375, opened 2021-06-09
apply ddl change for pqzip_url and fix lat/lon type change

bpasquer:


api/src/main/resources/sql/application.sql
```diff
@@ -132,14 +132,16 @@ CREATE TABLE nrmn.survey_aud (
     direction_mod boolean,
     inside_marine_park varchar(255),
     inside_marine_park_mod boolean,
-    latitude float8,
+    latitude float4,
```
jonescc, line 135: These should match (be generated from) from entity definitions in the java code.   I'll fix those.

api/src/main/resources/sql/application.sql
```diff
@@ -132,14 +132,16 @@ CREATE TABLE nrmn.survey_aud (
...
     latitude_mod boolean,
-    longitude float8,
+    longitude float4,
     longitude_mod boolean,
     notes varchar(255),
     notes_mod boolean,
     pq_catalogued boolean,
     pq_catalogued_mod boolean,
+    pq_zip_url varchar(255),
```
jonescc, line 143: Also needs to be added to the entity defns otherwise this won't get updated.

api/src/main/resources/sql/migration.sql
```diff
@@ -549,6 +549,7 @@ CREATE TABLE nrmn.survey (
     inside_marine_park character varying(50),
     notes character varying(1000),
     pq_catalogued boolean,
+    pq_zip_url varchar(255),
```
jonescc, line 552: Does this need to be displayed/editable in the app?

ccmoloney:
It might need to be displayed.  We probably have some plumbing to do...

api/src/main/resources/sql/application.sql
```diff
@@ -132,14 +132,16 @@ CREATE TABLE nrmn.survey_aud (
...
+    pq_zip_url_aud boolean,
```
jonescc, line 144: Needs to be _mod.

jonescc: Fixing...

jonescc (commented):
Aligned entity definitions with ddl and changed pq_zip_url_aud to pq_zip_url_mod.

PR can be merged but will need further changes to display pq_zip_url in the front end as required.

ccmoloney (approved):

PR #391, opened 2021-06-16
Default survey num to 0

ccmoloney:


api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/RawValidation.java
```diff
@@ -131,7 +131,7 @@ public StagedRowFormatted toFormat(HashMap<String, Object> values, boolean isExt
                 val splitDepth = values.get("Depth").orElseGet(null).toString().split("\\.");
                 val depth = Integer.parseInt(splitDepth[0]);
 
-                Optional<Integer> survey_num = splitDepth.length == 1 ? Optional.empty()
+                Optional<Integer> survey_num = splitDepth.length == 1 ? Optional.of(new Integer(0))
```
jonescc, line 134: Yep this will work - would be good to make this an Integer now instead of an Optional Integer..

PR #394, opened 2021-06-17
Add species name to l5/l95 warning message

ccmoloney:


jonescc:
image

jonescc:
Looks like ui_species_attribute has no null species_name - so some code problem.

jonescc:
Do we need to add this for LMax check as well?

PR #397, opened 2021-06-21
aodn/backlog#3126 - include missing properties in ingest

jonescc:
Also:

* work to double precision internally to minimise conversion issues.
* remove unused coordinates check covered by within 200m site check


jonescc:
Investigating an issue with longitude being rounded.

nspool (approved):

PR #406, opened 2021-06-25
Fix missing Species Epithet from Observable Item views

nspool:
Add species epithet to ObservableItemGetDto and also update API tests.


jonescc:
Need to fix conflict now after I merged the other PR

PR #408, opened 2021-06-25
Datasheet Reimplementation

nspool:
- Reimplement the ag-grid to use an immutable data store. Store all grid metadata in the ag-grid Context property.
- Implement a bespoke Undo function to support row add and delete as well and find and replace.
- Change grid styling to use ag-grid cell styles instead of custom cell types.
- Reduce the use of material-ui Grid for layout and use flexbox driven Box to prevent.
- Rewrite find & replace and validation display to use the ag-grid ToolPanel.
- Replace the Staged Job Rows Post / Delete API methods with a single Put method to perform the grid save operation in a single transaction.
- Remove custom styling and use material-ui themes exclusively.
- Upgrade npm dependencies, React from 16 to 17 and Ag-grid from 23 to 25

Stack deployed to http://nrmn-reimplement-datasheet.dev.aodn.org.au


api/src/main/java/au/org/aodn/nrmn/restapi/service/StagedRowService.java
```diff
@@ -20,19 +22,15 @@
...
-    }
-
-    public Validated<ErrorInput, Integer> delete( List<StagedRow> toDeleteRows) {
-        rowRepo.deleteAll(toDeleteRows);
-        rowRepo.flush();
-        return Validated.<ErrorInput, Integer>valid(toDeleteRows.size());
+    @Transactional
+    public Validated<ErrorInput, Integer> save(Long jobId, List<Long> toDeleteRowIds, List<StagedRow> newRows) {
+        rowRepo.deleteAllByIds(toDeleteRowIds);
+        Integer rowsAdded = 0;
```
jonescc, line 28: can't rows be updated as well?

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyContentsHandler.java
```diff
@@ -45,7 +45,7 @@ public void startRow(int rowNum) {
         isDataRow = (rowNum > 1);
         rowHasData = false;
         if (isDataRow) {
-            currentRow = StagedRow.builder().pos(rowNum - 1).build();
+            currentRow = StagedRow.builder().pos((rowNum - 1) * 1000).build();
```
jonescc, line 48: Whats this magic number?

api/src/main/java/au/org/aodn/nrmn/restapi/service/StagedRowService.java
nspool: ATM the client only cares if rows are added so that it can perform a full reload in order to get the server-assigned row IDs.

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyContentsHandler.java
nspool: Some slack so that the client can increment 1 to the row number when cloning a row and still maintain row order.

api/src/main/java/au/org/aodn/nrmn/restapi/service/StagedRowService.java
jonescc: OK that sounds like it might result  in weird behaviour where you refresh and edits are gone but deletes and inserts saved.  For later then.

nspool:
Changed to draft to fix issue were grid loses formatting when menu opened.

PR #413, opened 2021-06-30
Check for duplicate rows

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3168


jonescc:
I'm getting a duplicate row warning for 2019AutumnADP_RLSsurveys_edit4.xlsx which doesn't look right:

image

In the source spreadsheet depth is different.  Note that depth is a composite field and is split into depth and surveyNum in stagedrowformatted.

APaulic:
> I'm getting a duplicate row warning for 2019AutumnADP_RLSsurveys_edit4.xlsx which doesn't look right:
> 
> image
> 
> In the source spreadsheet depth is different. Note that depth is a composite field and is split into depth and surveyNum in stagedrowformatted.

@jonescc Oh, I didn't know that. I have just updated it to check the split column correctly.

PR #422, opened 2021-07-04
Include rowId in global error messages

APaulic:
Improvement of work in https://github.com/aodn/backlog/issues/3168

@nspool This will make the Duplicate Rows Check include the rowId in the payload. This is the first step in highlighting the duplicated row. It requires front-end changes to match.

Global Error payload:
```json
{
   "errorGlobal":[
      {
         "rowId":1500,
         "jobId":14,
         "message":"Row 99 may be a duplicate of row 98.",
         "errorType":"GLOBAL",
         "errorLevel":"WARNING",
         "columnTarget":"Duplicates"
      }
   ]
}
```

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/global/formatted/DuplicateRowCheck.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/global/formatted/DuplicateRowCheck.java
@@ -12,6 +12,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 public class DuplicateRowCheck extends BaseGlobalFormattedValidator {
+
     public DuplicateRowCheck() {
         super("Duplicates");
     }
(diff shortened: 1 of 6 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/validation/validators/global/formatted/DuplicateRowCheck.java
```diff
@@ -36,10 +37,9 @@ public DuplicateRowCheck() {
                 .map(duplicateRow ->
                         invalid(
                                 job.getId(),
-                                String.format("Row %s may be a duplicate of row %s.",
-                                        duplicateRow.getRef().getPos(),
-                                        duplicateRow.getRef().getPos() - 1),
-                                ValidationLevel.WARNING)
+                                "Row may be a duplicate of preceding row",
```
nspool, line 40: Remove 'of preceding row' stil duplicates should be detected even if they are not grouped together.

APaulic: The logic only finds duplicates that immediately follow the original row

nspool:
@APaulic front-end highlighting has been implemented. Clicking on the warning should scroll to the point in the sheet with the duplicate row and highlight it.

PR #423, opened 2021-07-05
Datasheet fixes

nspool:
- Find & replace is not using column display order and is missing option for substring matching
- Undoing a cloned row is not removing the added row on undo
- Fix tool panel css so that it is wider by default


nspool:
Converting to draft to investigate why sometimes the find result is not being focused.

PR #424, opened 2021-07-06
Show affected rows for global validations

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3203


nspool (commented):
Looks good. Replacing those tuples makes it much more readable.

PR #440, opened 2021-07-15
Delete multiple rows

APaulic:
I've just replicated the delete functionality to multiple rows which appears to work even with filters. I haven't been able to find an issue after testing it so I am quite confident there shouldn't be any side effects.


nspool (approved):
Works pretty well.

PR #444, opened 2021-07-19
Update form with seleted species

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3139


nspool (approved):
This will make the field a controlled component which can have performance implications.

PR #446, opened 2021-07-20
Replace timeline with table

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3113

Full/Large screen:
image

Half screen:
image


ccmoloney:
Can we also replace the 'tic tacs' with a simple tablular display of the job properties?  (Program, Status, Job Tupe, Extended Size, Created date, User, etc...)

nspool (approved):

PR #453, opened 2021-07-27
Include diver names in survey list (Requires Optimisation)

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3294


APaulic:
The main performance issue is from the fact that it is a 9MB payload. In order to make it faster, it would need to be paged.

nspool (approved):

nspool:
@APaulic I think paging is something we'll need to introduce in day 2.

PR #460, opened 2021-07-28
Fix incompatible reactor-core + netty versions

nspool:
Versions need to match up to what is displayed on the release page at https://github.com/reactor/reactor/releases.


nspool:
Converting back to draft while checking for other issues.

nspool:
S3 upload doesn't appear to be implemented, so the new code is  failing with an Access Denied. Will comment out for now.

APaulic (approved):

PR #465, opened 2021-07-29
AG Grid Config Endpoint

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3295


ui/src/components/import/DataSheetView.js
```diff
@@ -392,6 +393,13 @@ const DataSheetView = ({jobId, onIngest}) => {
 
   const onGridReady = (p) => {
     setGridApi(p.api);
+
+    getResult('config/aggrid').then((result) => {
```
nspool, line 397: It could be neater to return the ag-grid key along with the JWT on login. But then we'd need to persist it locally in a way that wasn't obvious.

nspool (approved):

nspool (changes requested):
I think `aggrid.license=` need to be added to the test application.properties

APaulic:
> I think `aggrid.license=` needs to be added to the test application.properties

I just moved the import which I believe makes it lazy

PR #511, opened 2021-08-05
Stage only rows with IDs

nspool:
Also log the number of rows skipped that contain data

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/controller/StagedJobController.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/controller/StagedJobController.java
@@ -138,9 +138,9 @@ public class StagedJobController {
             jobRepo.save(job);
 
             List<StagedRow> rowsToSave = parsedSheet.getStagedRows();
-            Long numEmptyRows = parsedSheet.getNumEmptyRows();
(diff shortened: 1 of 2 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/controller/StagedJobController.java
```diff
@@ -138,9 +138,9 @@
             jobRepo.save(job);
 
             List<StagedRow> rowsToSave = parsedSheet.getStagedRows();
-            Long numEmptyRows = parsedSheet.getNumEmptyRows();
+            Long skippedRows = parsedSheet.getSkippedRows();
 
-            String message = "Staged " + (rowsToSave.size()) + " row(s)." + (numEmptyRows > 0 ? " " + numEmptyRows + " empty " + "row(s) skipped." : "");
+            String message = "Staged " + (rowsToSave.size()) + " row(s)." + (skippedRows > 0 ? " " + skippedRows + " rows missing and ID but containing data were skipped." : "");
```
ccmoloney, line 143: "missing and ID" should be "missing an ID"?

nspool: Fixed.

ccmoloney (approved):

PR #530, opened 2021-08-10
Update Tests

nspool:
Update all tests to use the new validation methods in the ValidationProcess classs.


nspool:
Removed some additional refactoring as it appears to have broken the test config.

PR #537, opened 2021-08-10
Validation improvements

nspool:
Species: skip superseded validation if superseded value is am empty string
Duplicate rows: improve row hashing

Also fix a potential issue where the validation panel summary may not be refreshed on Save & Validate.


api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/ValidationProcess.java
```diff
@@ -153,7 +149,7 @@ private String hashRow(StagedRow r) {
                 Optional<ObservableItem> observableItem = species.stream().filter(s -> row.getSpecies().equalsIgnoreCase(s.getObservableItemName())).findAny();
                 if (observableItem.isPresent()) {
                     String supersededBy = observableItem.get().getSupersededBy();
-                    if (supersededBy != null)
+                    if (supersededBy != null && supersededBy.length() > 0)
```
ccmoloney, line 152: worth trimming before checking length?

nspool: PR updated.

PR #551, opened 2021-08-12
Ingest - Null check for direction + fix vis as double

APaulic:
Backlog item: https://github.com/aodn/backlog/issues/3361


nspool (approved):

ccmoloney:
Verified in edge.  DODed.

PR #571, opened 2021-08-18
Fix summary counts

nspool:
Site & Diver count were not being done case-insensitive


nspool:
Back to draft while investigating distinct survey count.

nspool:
Fix regression in distinct survey count introduced in a previous fixed.

APaulic (approved):

PR #617, opened 2021-09-09
Remember datasheet filter and position on save

nspool:
Do not reset the filter or sheet position on save.
Also fix the columns resizing to minimum width after a full reload.


nspool:
@APaulic demo available to http://nrmn-aodnstack-nspool.dev.aodn.org.au/

APaulic (approved):

PR #639, opened 2021-09-22
Use bump2version to bump versions instead of maven plugin

akashisama:


nspool:
@akashisama Merged but the version stopped bumping and was stuck at 0.0.0. Reverting for now.

PR #640, opened 2021-09-22
Replace public views: exclude A ATRC surveys; remove wave_exposure an…

bpasquer:
…d protection_status from site and survey views


nspool:
Ran on nrmn_dev, nrmn_edge and nrmn_prod.

PR #662, opened 2021-10-05
Drop site_list view + condition based on survey_id

bpasquer:


nspool:
@bpasquer on line 248 the column "method" doesn't appear to exist on `ep_m0_off_transect_sighting`.

bpasquer:
> @bpasquer on line 248 the column "method" doesn't appear to exist on `ep_m0_off_transect_sighting`.

That's a change I'll have to do to the private endpoint first in a separate PR.

scripts/db-update/06-update_defintion_public_views.sql
```diff
@@ -244,19 +245,16 @@ SELECT
...
 	phylum,
 	"class",
 	"order",
 	family,
 	species_name,
-	reporting_name,
-	size_class,
-    total,
-    biomass
+	reporting_name,,
```
nspool, line 254: Is this extra comma a typo?

nspool (changes requested):

PR #763, opened 2022-02-08
mapped_id added to ep_obs_item and RLS webstite endpoints

bpasquer:


db/endpoints/CreatePrivateEndpoints.sql
```diff
@@ -1085,7 +1087,8 @@ select epoi.observable_item_id as species_id,
 	common_phylum_name,
 	null as geom,
 	epoi.superseded_ids,
-	epoi.superseded_names from nrmn.ep_observable_items epoi
+	epoi.superseded_names from nrmn.ep_observable_items epoi,
+	epoi.mapped_id
```
nspool, line 1091: I think this should be on line 1090 before  `from nrmn.ep_observable_items epoi`

nspool (changes requested):

db/endpoints/CreatePrivateEndpoints.sql
bpasquer: indeed!

nspool (approved):

PR #776, opened 2022-03-03
Improve API Code Quality

nspool:
Removing unused classes and dead code and upgrading deps.
Add explicit repositories, controllers and DTOs for all reference data metadata (eg. taxnomy, locations, programs etc).
Change the Diver list so that it can be edited from the grid.
Reformat code that was using 8 space indentation instead of 4.

Stack available at: http://nrmn-aodnstack-nsp.dev.aodn.org.au/home


gsatimos:
whoa that's a lot of files. Might need you to take me through this @nspool . Also the stack you mentioned isn't running right now

gsatimos:
Not sure if regression or just separate bug but there's no feedback on the form when entering initials for a diver where the initials already exists:

image

nspool:
@gsatimos this is an existing bug.

nspool:
@gsatimos fixed error message not appearing. stack is back up at http://nrmn-aodnstack-nsp.dev.aodn.org.au/home.

gsatimos (approved):

PR #824, opened 2022-03-15
Update high level description

craigrose:


gsatimos:
you prob don't need to go into that much detail in reviewing the README. As long as the title there is a sentence or two after the title briefly describing what the app/tool/whatever is for then that would be enough

gsatimos (approved):

PR #895, opened 2022-04-03
Improve ingest feedback

nspool:
- Display ingest errors on the datasheet screen, otherwise redirect to the job on successful ingestion.
- Update FILL_DATA.sql to use the same size classes and measurement names as production.


utas-raymondng (approved):
LGTM

PR #903, opened 2022-04-05
Put back survey distance from site

nspool:
Survey validation should be displaying the distance from the site if greater than 10m.

I accidentally remove this from the message in the last PR.


utas-raymondng (approved):
LGTM

PR #906, opened 2022-04-06
Template Export: Filter by method before generating letter codes

nspool:
Generate the letter code map on a per-method basis to fix long letter codes.

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/service/TemplateService.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/service/TemplateService.java
@@ -54,7 +54,7 @@ public class TemplateService {
     ObservableItemRepository observableItemRepository;
 
     @Autowired
-    LetterCodeRepository LetterCodeRepository;
+    LetterCodeRepository letterCodeRepository;
 
     @Autowired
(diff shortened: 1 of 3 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/service/TemplateService.java
```diff
@@ -173,8 +170,12 @@ public void writeSpeciesCsv(Writer writer, CSVFormat csvFormat, Collection<Speci
                 toString(species.getLMax()));
     }
 
-    public List<SpeciesWithAttributesCsvRow> getSpeciesForTemplate(Integer mode, Collection<Integer> siteIds,
-            HashMap<Long, String> letterCodeMap) {
+    public List<SpeciesWithAttributesCsvRow> getSpeciesForTemplate(Integer mode, List<Integer> siteIds) {
+
+        HashMap<Long, String> letterCodeMap = new HashMap<Long, String>();
```
utas-raymondng, line 175: You do not need to write the type here, which is less verbose.

HashMap<Long, String> letterCodeMap = new HashMap<>();

utas-raymondng (approved):
Minor comment, not affecting overall code quality.

PR #922, opened 2022-04-10
Use TypeScript for UI Tests

nspool:
Specifically so that mocked web requests and responses can be typed to match Spring Boot as closely as possible.


utas-raymondng (approved):
LGTM

PR #966, opened 2022-04-26
Ag Grid: Revert to version 27.1

nspool:
27.2 appears to have a breaking change preventing grids from rendering.


utas-raymondng (approved):
LGTM

PR #982, opened 2022-05-08
Add prefix to invert measurement validations

nspool:
Adding the prefix 'invert' will cause all invert measurement validations to be grouped as they are displayed alphabetically.


utas-raymondng (approved):
LGTM

PR #984, opened 2022-05-09
Print survey with decimal depth in validation messages

nspool:
Format the survey as "[site, date, depth.surveyNumber]" in the validation messages instead of just printing "site/date/depth".
This adds extra space to make it easier to read and also shows the survey number.

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/ValidationProcess.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/ValidationProcess.java
@@ -396,7 +396,7 @@ public class ValidationProcess {
 
             if (!existingSurveys.isEmpty()) {
                 Survey existingSurvey = existingSurveys.stream().findFirst().get();
-                String message = "Survey exists: " + existingSurvey.getSurveyId() + " includes [" + row.getSite().getSiteCode() + ", " + row.getDate() + ", " + row.getDepth() + "]";
+                String message = "Survey exists: " + existingSurvey.getSurveyId() + " includes " +row.getDecimalSurvey();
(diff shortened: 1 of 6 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/ValidationProcess.java
```diff
@@ -465,19 +465,19 @@ public ValidationError validateMethod3QuadratsGT50(String transect, List<StagedR
...
     }
 
-    private ValidationError validateSurveyComplete(String programName, String surveyKey, List<StagedRowFormatted> surveyRows) {
+    private ValidationError validateSurveyComplete(String programName, List<StagedRowFormatted> surveyRows) {
 
         if(surveyRows.stream().anyMatch(r -> r.getMethod() == null || r.getBlock() == null))
             return null;
 
-        String messagePrefix = "Survey incomplete: " + surveyKey + " ";
+        String messagePrefix = "Survey incomplete: " + surveyRows.get(0).getDecimalSurvey();
```
utas-raymondng, line 481: Is get(0) always have value that it will not result in IndexOutOfBoundsException?

utas-raymondng (commented):
Minor comment, if no issue I can approve it.

api/src/main/java/au/org/aodn/nrmn/restapi/validation/process/ValidationProcess.java
nspool: The aggregate on line ~739 creates a map with the survey rows, which makes sure there is at least one row for each survey. Line 564 also gets the first row by index.

utas-raymondng (approved):
LGTM

PR #998, opened 2022-05-16
Add back missing icons

utas-raymondng:
Add missing icons in site list, there is a problem with the delete icon and will be address with different ticket 

https://github.com/aodn/backlog/issues/4036

### DIFF
--- a/web/src/components/data-entities/site/SiteList.js
+++ b/web/src/components/data-entities/site/SiteList.js
@@ -142,7 +142,7 @@ const SiteList = () => {
             filter={false}
             resizable={false}
             sortable={false}
-            cellRenderer={() => <Delete/>}
+            cellRenderer={(e) => e.data.isActive ? <></> : <Delete/>}
             cellStyle={{paddingLeft: '10px', color: 'grey', cursor: 'pointer'}}
             onCellClicked={(e) => {
               setDialogState({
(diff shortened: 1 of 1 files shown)


web/src/components/data-entities/site/SiteList.js
```diff
@@ -142,9 +142,7 @@ const SiteList = () => {
             filter={false}
             resizable={false}
             sortable={false}
-            valueFormatter={(e) => {
-              return e.data.isActive ? '' : 'Delete';
-            }}
+            cellRenderer={() => <Delete/>}
```
nspool, line 145: This should probably be ` cellRenderer={(e) => e.data.isActive ? <></> : <Delete/>}`

nspool (changes requested):
Note that `CellRenderer` is much slower than `ValueRenderer` and can affect scrolling performance: https://www.ag-grid.com/javascript-data-grid/scrolling-performance/#:~:text=Check%20Cell%20Renderers,are%20adding%20the%20most%20overhead

It might be more performant to just to use a `ValueRenderer` with emoji.

web/src/components/data-entities/site/SiteList.js
utas-raymondng: updated

nspool (approved):

PR #1014, opened 2022-05-17
Site List: Disable delete handler for active sites

nspool:
Prevents the delete dialog from appearing when clicking an active site row in the 'delete' column. Also adds a Jest test for SiteList.


utas-raymondng (approved):
LGTM

PR #1015, opened 2022-05-17
add missing icon

utas-raymondng:


nspool (changes requested):
Please also match the gray colour for the grid icon, and also the spacing in the original screenshot. I think the grid might need to be in it's own column, and the row height increased.

utas-raymondng:
> Please also match the gray colour for the grid icon, and also the spacing in the original screenshot. I think the grid might need to be in it's own column, and the row height increased.

Updated as per comments

web/src/components/job/JobList.js
```diff
@@ -90,16 +99,17 @@ const JobList = () => {
               }
             }}
           />
+          <AgGridColumn
+            width={40}
+            cellStyle={{cursor: 'pointer'}}
+            cellRenderer={(e) => (e.data?.status === 'STAGED' ? <GridOn htmlColor={'#808080'} viewBox={iconViewBoxDimension}/> : <></>)}
```
nspool, line 105: Try and avoid inline styling like this.

web/src/components/job/JobList.js
```diff
@@ -21,6 +22,14 @@ const JobList = () => {
   const [rowData, setRowData] = useState([]);
   const [redirect, setRedirect] = useState();
   const [deleteJobId, setDeleteJobId] = useState();
+  const iconViewBoxDimension = '-2 -2 30 30';
```
nspool, line 25: Try to avoid using transforms, especially for elements in grids where there are many rows.

nspool (approved):

PR #1016, opened 2022-05-17
Add version number to login box

utas-raymondng:


nspool (approved):
Moved the version to underneath the form.

PR #1021, opened 2022-05-20
Add table styles and apply to species search to improve readability

nspool:
Add styles to the global theme and applies them to the Species Search table.
Also fixes console warning due to use of `min-width`. Replaces with the AgGrid column prop `minWidth`.


utas-raymondng (approved):
LGTM

PR #1024, opened 2022-05-23
Add Unicode byte order mark to template export diver CSV

nspool:
Some versions of Excel apparently require this regardless of encoding. See: https://stackoverflow.com/questions/155097/microsoft-excel-mangles-diacritics-in-csv-files


utas-raymondng (approved):
LGTM

PR #1029, opened 2022-05-23
3823 match import export header

utas-raymondng:


api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
```diff
@@ -125,6 +124,18 @@ public void shortInvertsRowShouldBeInvalid() throws Exception {
...
         assertTrue(StringUtils.isNotEmpty(error));
     }
+    /**
+     * Excel row2 must be blank according to the import file spec, if not blank throw exception
+     */
+    @Test
+    void rejectFileWhenRow2IsNotEmpty() {
+        Exception e = assertThrows(Exception.class, () -> {
+            FileSystemResource file3 = new FileSystemResource("src/test/resources/sheets/row2NotEmpty.xlsx");
+            ParsedSheet parsedSheet = sheetService
```
nspool, line 134: Unused variable?

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyContentsHandler.java
```diff
@@ -14,8 +15,8 @@
 
 public class SurveyContentsHandler implements SheetContentsHandler {
 
-    private final List<String> header1Required;
-    private final List<String> header1Optional;
+    private final List<Field> header1Required;
```
nspool, line 18: `Field` is a bit too general, could it be renamed `SurveyField` instead?

nspool, line 18: `Field` is a bit too general, could it be renamed `SurveyField` instead?

nspool (changes requested):
It should be possible to add the blank row using the `prependContent` property. See: https://www.ag-grid.com/react-data-grid/excel-export-extra-content/

api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
utas-raymondng: Yes, removed as the code expected exception

utas-raymondng:
> It should be possible to add the blank row using the `prependContent` property. See: https://www.ag-grid.com/react-data-grid/excel-export-extra-content/

Correct, implemented

Regarding the ; at the bottom of the page, it is due to ; appears in the ValidationJob.js. I fixed it together in this change.

nspool (changes requested):
Getting this error when trying to reingest an exported sheet: `Row 1 missing headers: 50. Row 1 has unexpected headers: UNKNOWN`.

Please also move the Field enum into a separate file and rename it something like `SurveyField`.

api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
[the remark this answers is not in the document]
utas-raymondng: > Getting this error when trying to reingest an exported sheet: `Row 1 missing headers: 50. Row 1 has unexpected headers: UNKNOWN`.
> 
> Please also move the Field enum into a separate file and rename it something like `SurveyField`.

The missing headers error is due to a mismatch type in constant.js, I fixed it and make a conversion in export function so that the header will always be type 'String' 

SurveyField refactored and become a standalone file

nspool (approved):

PR #1044, opened 2022-05-26
Add Jest test for Species Search

nspool:


web/src/components/import/__test__/DataSheetView.test.tsx
```diff
@@ -1,4 +1,3 @@
-// @ts-ignore
```
nspool, line 1: removed some unneeded things

web/src/components/search/SpeciesSearch.js
```diff
@@ -21,20 +21,20 @@ const useStyles = makeStyles(({ palette, typography }) => ({
       '& .MuiTableHead-root': {
         '& .MuiTableRow-head': {
           '& .MuiTableCell-head': {
-            fontSize: typography.table.fontSize,
```
nspool, line 24: needed for tests since we aren't applying a theme and don't care about styling for the tests.

craigrose (approved):

PR #1045, opened 2022-05-30
Fix Species Search pagination

nspool:
- Fix bug causing pagination to break when last page is less than maxPage size
- Update test to check the count in the footer (eg `1-50`) are correct


utas-raymondng (approved):
LGTM

PR #1046, opened 2022-05-30
4044 persist filters

utas-raymondng:


nspool (changes requested):
Instead of removing TypeScript we should work out how to get IntelliJ to support it. Also
- Please remove package-lock.json as it is not necessary for Yarn.
- AgGridColumn `headerName` props are not necessary.
- It would be better to give the sample data files a more descriptive name ie 'JobList.data.json'

utas-raymondng:
> Instead of removing TypeScript we should work out how to get IntelliJ to support it. Also
> 
>     * Please remove package-lock.json as it is not necessary for Yarn.
> 
>     * AgGridColumn `headerName` props are not necessary.
> 
>     * It would be better to give the sample data files a more descriptive name ie 'JobList.data.json'

1. package-lock removed
2. The reason I added the headerName is because the headerName is generated from the field name, however the test case verify the header name exist. I want to have a fix string to compare instead of using a name generated by ag-grid 
3. Sample data name changed

nspool:
Thanks. Use the selector `.ag-cell[col-id="columnName"]` to get the column instead of setting the `headerName` prop and querying for the full header text.

utas-raymondng:
> .ag-cell[col-id="columnName"]

updated to use col-id

nspool (approved):

PR #1057, opened 2022-05-31
Improve SiteList Test

nspool:
Looping on the expect within await appears to sometimes fail. Fix so the await is in the innermost loop.

Also remove some unused imports and unneeded annotations.


utas-raymondng (approved):
LGTM

PR #1061, opened 2022-06-02
3445 warn new habitat group

utas-raymondng:


web/src/components/input/CustomAutoCompleteInput.js
```diff
@@ -1,21 +1,93 @@
...
 import {PropTypes} from 'prop-types';
+import { makeStyles } from '@mui/styles';
 
-const CustomAutoCompleteInput = ({label, field, options, onChange, formData, errors}) => {
-  const error = errors?.find((f) => f.property === field);
+export const ERROR_TYPE = {NORMAL: 0, WARNING: 1, ERROR: 2};
+
+const CustomAutoCompleteInput = ({label, field, options, onChange, formData, errors, warnLevelOnNewValue = ERROR_TYPE.NORMAL}) => {
+
+  const useStyles = value =>
```
nspool, line 11: Move the makeStyles outside of the component since it will be called every time the it's rendered.

nspool (changes requested):
Works well! Just move the makeStyles outside of the component since it will be called every time the it's rendered.

web/src/components/input/CustomAutoCompleteInput.js
utas-raymondng: Moved, please wait for rebuild to be completed.

PR #1078, opened 2022-06-09
Fix potential ReDoS in WoRMS search regex

nspool:
Split `(` and `[` checks into different capturing groups.

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/service/WormsService.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/service/WormsService.java
@@ -21,7 +21,7 @@ import java.util.stream.Collectors;
 @Service
 public class WormsService {
 
-    static private Pattern REMOVE_TRAILING_JUNK_PATTERN = Pattern.compile("( (sp\\.|spp\\.))?( (\\(.*\\)|\\[.*\\]))?$",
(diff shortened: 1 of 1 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/service/WormsService.java
```diff
@@ -21,7 +21,7 @@
 @Service
 public class WormsService {
 
-    static private Pattern REMOVE_TRAILING_JUNK_PATTERN = Pattern.compile("( (sp\\.|spp\\.))?( (\\(.*\\)|\\[.*\\]))?$",
+    static private Pattern REMOVE_TRAILING_JUNK_PATTERN = Pattern.compile("(( (sp\\.|spp\\.).+)|( \\(.+)|( \\[.+))",
```
utas-raymondng, line 24: Can you add a test case for this reg expression as it is hard to verify later what you want to filter out what not.

nspool: @utas-raymondng here is it: https://github.com/aodn/nrmn-application/blob/master/api/src/test/java/au/org/aodn/nrmn/restapi/service/WormsServiceTest.java

utas-raymondng (approved):
LGTM

PR #1089, opened 2022-07-03
Survey corrections service

nspool:
Survey measurements can now be added, updated or deleted and surveys can be removed. All correction operations will create a Job of type "Corrections" containing a summary of operations performed.

All functions should be gated behind the "Corrections" feature flag.

- Add a CorrectionsService class to delete existing and add new observations to surveys that are being corrected, as well as delete all observations associated with a survey as well as the survey itself. 
- Implement different corrections validation checks, including a reimplementation of the format validation which works by mapping back the `StagedRowFormatted` to the `StagedRow` and so can verify correct formatting by comparing the original row with the mapped row.
- Remove the ValidationConstraintService since I don't think it is useful in practice. Instead, fields that can not be corrected are disabled in the AG Grid column definition.
- Implement the corrections UI workflow by adding a summary panel and also a confirmation step which is designed to eventually show a summary of what operations will be performed immediately before confirming.


utas-raymondng (commented):
It looks like that even user change 1 row in the survey, all rows in the survey get re-created. There may be problem if two people loaded the same survey and edit different row but save it at different time.

nspool:
@utas-raymondng Yes that's right. The whole correction update will complete in a transaction so that if another update is performed while one is still in progress, it will fail. This is how the ingest process works currently.

utas-raymondng:
> @utas-raymondng Yes that's right. The whole correction update will complete in a transaction so that if another update is performed while one is still in progress, it will fail. This is how the ingest process works currently.

What i mean is 
1. User A open the survey and edit row 1 value from a -> b
2. User B open the same survey and edit row 2 value from c -> d
3. User A save the survey and completed the transaction, now the survey will have row 1 value b and row 2 value c
4. User B then save the survey, since the user have not refresh the page, then row 1 value is still a on the user screen. After the save the survey will have row 1 value a and row 2 value d, which discarded user a change. 

Is this expected?

bpasquer:
> There may be problem if two people loaded the same survey and edit different row but save it at different time.
@utas-raymondng there is very small risk that 2 people work on the same survey as Toni and Lizzi are responsible for managing different datasets

nspool:
@utas-raymondng yes this is expected since this is a limitation of the existing ingestion process. There is a backlog item to implement optimistic locking in the future: https://github.com/aodn/backlog/issues/2424.

utas-raymondng:
> > There may be problem if two people loaded the same survey and edit different row but save it at different time.
> > @utas-raymondng there is very small risk that 2 people work on the same survey as Toni and Lizzi are responsible for managing different datasets

Ok if is expected behavior

utas-raymondng (approved):
LGTM

PR #1091, opened 2022-07-04
remove deprecated class

utas-raymondng:
Looks like the current master failed too when running on my PC, the branch build 1212-WORM-search build is fine so I believe one it is merged to master, this build should work too


nspool (approved):
LGTM

PR #1093, opened 2022-07-07
Increase font size of Species Search results

nspool:
Feedback indicated that the font size is too small.


utas-raymondng (approved):
LGTM

PR #1094, opened 2022-07-07
Add missing integer mappings for Block and Total columns

nspool:
Block and Total values are using the default modelmapper mapping resulting in an unhandled exception when either value is not a number.

Fix found another bug where the Latitude and Longitude values were being returned as NaN instead of null if the conversion failed.

Add an invalid mappings test.


utas-raymondng (approved):
LGTM

PR #1097, opened 2022-07-13
Speed up Survey and Site list queries

nspool:
- Rewrite the Survey List native query as two cacheable HQL queries.
- Add missing Cacheable annotation to the Site `getAll` method.

Survey List query should now complete in about a second, down from ~10 seconds.


utas-raymondng (approved):
You may consider execute those find function on application start so that the first user do not need to wait.

Other than that LGTM

PR #1100, opened 2022-07-27
Implement rate-limiting on auth endpoint methods

nspool:
- Implements rate-limiting using bucket4j.
- Allow 20 auth requests every 10 minutes per IP address.
- When limit is reached display how long to wait until next attempt.


utas-raymondng (approved):
LGTM

PR #1107, opened 2022-08-03
Rename correction job status from CORRECTION to CORRECTED

nspool:
Change correction status from  "CORRECTION" to "CORRECTED" so that it matches the other job status types.
Requires a database update to be run after merge.


utas-raymondng (approved):
LGTM

PR #1109, opened 2022-08-04
Record modified IDs as a range on the correction job

nspool:
Instead of printing every single observation ID the log will now only record a range (unless the range is not contiguous).


utas-raymondng (approved):
LGTM

PR #1112, opened 2022-08-11
4131 paginated survey

utas-raymondng:
Can you do a quick review and comment if you have concern on anything, especially I add the 1 to many mapping in Survey.class SurveyMethodEntity.class so that I can navigate from Survey to the Diver

I changed the grid UI to use setDatasource(), so it fetch data 100 by 100 from server.  The dynamicQuery is the key to achieve the complex query that generate from UI and execute on server. 

The diver filter is handle differently that if the filter exist, the system will query all unique survey id where it matches the diver filter and then set those survey id as a filter and pass it to the survey query.

In additional I added a PGDialect class where it extends the original use Dialect so that I can inject a custom function that concat diver name

Please do not merge at the moment, if you have no concern, I will add auto test case to make sure all sql works


nspool:
@utas-raymondng looks like a test is failing with `AG Grid: setFilterModel() - no column found for colId: siteName`

utas-raymondng:
> @utas-raymondng looks like a test is failing with `AG Grid: setFilterModel() - no column found for colId: siteName`

Yes there are few test build, I will fix it.

nspool (approved):

PR #1114, opened 2022-08-22
View source data sheet for ingested jobs

nspool:
- Show sheet icon in Job List for Ingested as well as Staged sheets
- Display original datasheet as read-only for ingested jobs
- Add to datasheet from Job View page
- Fix the S3 download link in the Job View page

Also remove some unused vars and fix compiler type warnings.


utas-raymondng (approved):
LGTM

PR #1126, opened 2022-09-07
Implement editing customisations for Corrections sheets

nspool:
Implement Find / Replace and Clone / Delete row and error highlighting as the same as the ingestion sheet.
Filter and go to value when clicking a validation.
Also update tests and VSCode launch configuration.


utas-raymondng (approved):
LGTM

PR #1127, opened 2022-09-07
Misc performance and code improvements

nspool:
Reference data grid: use onBodyScrollEnd instead of onBodyScroll so that the grid is only sized once at the end of a scroll.
API: remove unused imports, annotate some raw types and add default case to switch statements that are missing.
~Site Entity: use a Lombok default builder for setting default value of isActive.~ 
Reverted change to the site entity since the original behaviour is correct, even though there is a warning.


alexgleith (dismissed):
🚀

utas-raymondng (approved):
LGTM

PR #1131, opened 2022-09-27
Corrections measurement validation

nspool:
Refactors the following validations into `MeasurementValidation` service which is called by both the Ingest and Correction validation step:
-  Debris Zero observations
-  Record has no data and but not flagged as 'Survey Not Done' or 'No Species Found'
-  At least one value recorded in any of the size class columns or in the column Inverts
-  Abundance CheckSum is correct
-  Species size within L5 - L95
-  Species size below LMax
-  Species Abundance Check

Also adds a mapping for species letter code which was missing and meant that SND/DEZ rules were being ignored.

Move integration tests into their own folder and started grouping unit tests together under the service being tested (so far just 'measurements') and added missing tests for:
- Row has no data and but not flagged as 'Survey Not Done' or 'No Species Found'"
- 'Survey Not Done' has Value/Total/Inverts not 0 or 1
- (Invert) Measurements outside L5/95
- (Invert) Measurement above Lmax
- Row contains no measurements but is a measure method


utas-raymondng (approved):
LGTM

PR #1132, opened 2022-09-29
Corrections reference data and survey validations

nspool:
Implements the remaining validations needed for single-survey corrections.

Refactors validation methods from the `ValidationProcess` class into `SurveyValidation`, `SiteValidation` and `SpeciesFormatting` services.

---

Checks excluded from corrections:
- Survey Group (since this is single-survey only)
- Buddy (since buddy is not ingested and can't be corrected)
- Total (since the checksum is present for Excel sheets only)

Validations refactored and added to corrections:

Block checks
- M1, M2 (and M3 if ATRC) are present
- M1, M2 each have B1, B2 (and M3 has B0 if ATRC)
- M0 has B0, B1 or B2

Survey checks
- Survey Is New (and fail if survey ID doesn't match the survey being corrected
- Survey Group (available but disabled since it will always fail for a single survey)
- SurveyTransects (survey has 1,2,3,4 survey numbers)
- M3 checks that transect sums to 50 but M3 is less than 50
- ATRC / RLS Future Survey Rule / Surveys Too Old
- Survey not more than 10km from Site
 
Species checks
- Inverts are zero for M3, M4, M5
- Species are valid for method
- Abundance exceeds MaxAbundance
- Observations are under Lmax and within L5/95 
- 'Survey Not Done' / 'No Species Found' / Debris has Value/Total/Inverts of 0 or 1

Other checks
- Duplicate Rows check
- Row contains no measurements
- Row has no data and no value recorded for inverts
- Row has no data and but not flagged as 'Survey Not Done' or 'No Species Found'

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/controller/CorrectionController.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/controller/CorrectionController.java
@@ -208,7 +208,7 @@ public class CorrectionController {
             {
                 put("block", "Block is not an integer");
                 put("date", "Date format not valid");
-                put("depth", "Depth is not an interger");
+                put("depth", "Depth is not an integer");
                 put("direction", "Direction is not valid");
                 put("diver", "Diver does not exist");
                 put("latitude", "Latitude is not a decimal");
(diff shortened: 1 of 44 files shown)


utas-raymondng (approved):
LGTM

PR #1133, opened 2022-10-04
Fix 'Save' overlay not appearing after first save

nspool:
Fix 'Save' overlay not appearing after first save when editing and ingest sheet.


utas-raymondng (approved):
LGTM

PR #1137, opened 2022-10-12
Datasheet fixes and improvements

nspool:
Fixes:
- Edit Survey Date fails without error message
- Survey transect validation failing for surveys without numbers (eg http://nrmn-edge.aodn.org.au/data/survey/923401215)
- 'No Species Found' and 'Debris Zero' validation always failing
- Row number always showing as zero for cloned rows
- Row number not showing in corrections validation messages

Also:
- Adds a new menu option when right-clicking Species column to 'View on reeflifesurvey'
- Improvements performance when rendering validation results, esp. if there are duplicate rows
- Increases the timeout of Jest tests from 10s to 30s as some tests (eg. DiverListFilter) where borderline


utas-raymondng (changes requested):
Some bugs mentioned in the ticket

utas-raymondng (approved):
LGTM

PR #1138, opened 2022-10-12
Make sure exception appears in log for sumologic

utas-raymondng:
Found the following unchecked Exceptions:

au.org.aodn.nrmn.restapi.controller.exception.ResourceNotFoundException
au.org.aodn.nrmn.restapi.controller.exception.ValidationException
software.amazon.awssdk.core.exception.SdkClientException
software.amazon.awssdk.services.s3.model.NoSuchKeyException
org.springframework.security.core.userdetails.UsernameNotFoundException
java.time.format.DateTimeParseException
java.time.format.DateTimeParseException
au.org.aodn.nrmn.restapi.security.RequestRejectedException

A few of it will print log entry

| Log level | Message |
|------ | ------- |
| Error | Ingestion Failed |
| Error | Could not download original file for job %s. %s" | 
| Error |  key "%s" does not exist in bucket "%s". Could not download original file for job %s| 
| Error |  Could not set user authentication in security context|

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyEditService.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyEditService.java
@@ -58,7 +58,9 @@ public class SurveyEditService {
 
     public Survey updateSurvey(SurveyDto surveyDto) {
 
-        Survey survey = surveyRepository.findById(surveyDto.getSurveyId()).orElseThrow(ResourceNotFoundException::new);
+        Survey survey = surveyRepository
+                .findById(surveyDto.getSurveyId()).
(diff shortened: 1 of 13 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyEditService.java
```diff
@@ -58,7 +58,9 @@ public class SurveyEditService {
 
     public Survey updateSurvey(SurveyDto surveyDto) {
 
-        Survey survey = surveyRepository.findById(surveyDto.getSurveyId()).orElseThrow(ResourceNotFoundException::new);
+        Survey survey = surveyRepository
```
nspool, line 61: I think using `var` when possible makes the code look more consistent.

nspool (approved):
LGTM

PR #1141, opened 2022-10-23
Fix potential security issues

nspool:
Upgrade dependencies to fix these dependabot alerts:
 - https://github.com/aodn/nrmn-application/security/dependabot/39
 - https://github.com/aodn/nrmn-application/security/dependabot/40


alexgleith (approved):
Looks sensible to me 🚀

PR #1143, opened 2022-10-24
Survey Corrections Summary Feature

nspool:
Adds to the corrections confirmation page a list of every change made to the corrections sheet prior to submitting.

### DIFF
--- a/web/src/components/data-entities/survey/SurveyCorrect.js
+++ b/web/src/components/data-entities/survey/SurveyCorrect.js
@@ -32,7 +32,7 @@ const toolTipValueGetter = ({context, data, colDef}) => {
 };
 
 const removeNullProperties = (obj) => {
-  return Object.fromEntries(Object.entries(obj).filter((v) => v[1] !== ''));
+  return obj ? Object.fromEntries(Object.entries(obj).filter((v) => v[1] !== '')) : null;
 };
 
 const packedData = (api) => {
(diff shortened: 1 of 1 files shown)


utas-raymondng (changes requested):
When I build it locally using mvn clean install, it always failed at the FAIL src/components/data-entities/diver/__test__/DiverList.filter.test.tsx (25.36 s)

<DiverList/> filter testing › Render necessary fields with filter restored expect(jest.fn()).toHaveBeenCalledTimes(expected)
    Expected number of calls: 1
    Received number of calls: 2

that may be reason for timeout in other failed test in the same file. However if I run it using npm test, it works

I do notices that the the get being called twice and hence the expected() failed. You can put a console.log in 

    // Override function so that it return the data we set.
    mockGetResult.mockImplementation((url) => {
      requestURL = url;

      const raw = {
        config: undefined,
        data: canned,
        headers: { 'Content-Type': 'application/json', 'Accept': 'application/json' },
        status: 200,
        statusText: url
      };

      return new Promise<AxiosResponse>((resolve => {
        console.log('call call ' + url);
        resolve(raw);
      }));
    });

and you see it called twice

Console
    console.log
      call call data/surveys?page=0&sort=%5B%7B%22field%22%3A%22survey.surveyId%22%2C%22order%22%3A%22desc%22%7D%5D
      at src/components/data-entities/survey/__test__/SurveyList.filter.test.tsx:104:17
    console.log
      call call data/surveys?page=0&filters=%5B%7B%22field%22%3A%22survey.siteName%22%2C%22ops%22%3A%22contains%22%2C%22val%22%3A%22South%20East%20Apple%22%7D%5D&sort=%5B%7B%22field%22%3A%22survey.surveyId%22%2C%22order%22%3A%22desc%22%7D%5D
      at src/components/data-entities/survey/__test__/SurveyList.filter.test.tsx:104:17

so the expected may be 2 instead of 1

nspool:
@utas-raymondng since this PR isn't changing those tests at all, can you please raise it as a seperate issue instead?

utas-raymondng:
> @utas-raymondng since this PR isn't changing those tests at all, can you please raise it as a seperate issue instead?

Sure issue raised https://github.com/aodn/backlog/issues/4455

utas-raymondng (approved):
Raised another issue to fix it 
https://github.com/aodn/backlog/issues/4455

PR #1151, opened 2022-11-02
Species correction bounding box filter

nspool:
Use the coordinates from the survey if available, otherwise use the site.


utas-raymondng (approved):
LGTM

PR #1176, opened 2022-11-13
Survey corrections export

nspool:
- Export entire sheet using the button
- Export a region by Shift-clicking and right-click "Export & Delete"


utas-raymondng (changes requested):
The export excel contains a column call "Buddy" but it is empty, export via "Export & Delete" method

[export_selected.xlsx](https://github.com/aodn/nrmn-application/files/9998613/export_selected.xlsx)

nspool:
@utas-raymondng `buddy` is needed but optional for ingest validation. But since the buddy itself is not ingested it is exported as an empty column.

utas-raymondng (approved):
As Buddy is expect empty, approve this PR

PR #1177, opened 2022-11-14
Support Debris Method 12

nspool:
Allow M12 on ingest and corrections
Require M12 to be Debris recorded on block 2


utas-raymondng (approved):
LGTM

PR #1179, opened 2022-11-14
Support Historical Method 10

nspool:
- Check that M10 has Block 1 and Block 2
- When validating species treat all M1 species as valid for M10


utas-raymondng (approved):
LGTM, however these validation rule is hard to understand by reading code, is there some place like README that can document it?

nspool:
@utas-raymondng currently they're only documented in the backlog issue: https://github.com/aodn/backlog/issues/4472.

PR #1180, opened 2022-11-14
Fix datasheet undo after deleting multiple rows

nspool:
Fixes a bug where after deleting a block of rows undo was only putting back a single row.


utas-raymondng (approved):
LGTM

PR #1184, opened 2022-11-16
Group correction confirmation messages by group

nspool:
If the same change has been made to multiple contiguous rows then display just one message with a range as the row value instead of a different message for each change.

Also log JWT expired token at Info not Error so it isn't reported by SumoLogic.


utas-raymondng (approved):
LGTM

PR #1186, opened 2022-11-16
Reset schema-update

nspool:
- `application.sql` previously updated with schema changes
- Data changes have been applied to production as of release 11.0 (723)


utas-raymondng (approved):
LGTM

PR #1187, opened 2022-11-17
Improve Species Correction Filters

nspool:
- Broader location fields should limit more specific filter selection. For example, selecting EcoRegion will only display the Country and Areas associated with that EcoRegion.
- Locations displayed on the corrections pane should correspond to surveys displayed.
- Locations displayed on the corrections pane should be removable. Click on the search result again to reset if necessary.
- Correction will only apply to surveys in the locations appearing on the panel.
- Adds the ability to filter by polygons provided in a KML file. 
To use, log in to Google and go to https://www.google.com/maps/d/edit. Filter only supports closed polygonal areas (see the screenshot). Export to KML and tick the checkout `Export as KML instead of KMZ. Does not support all icons.`. Add the KML file into the filter by selecting `Upload KML File`. If it is working correctly each shape should appear in the dropdown menu that appears.
map

I've run up a stack containing this feature: http://nrmn-verification.dev.aodn.org.au/data/species

<details>
  <summary>Login Credentials</summary>

```
username: info@aodn.org.au
password: password
```

</details>


utas-raymondng (approved):
LGTM

PR #1189, opened 2022-11-22
Survey Data Lock

nspool:
- Add a boolean `locked` column to the `survey` table that if true will prevent any corrections being performed on survey data as well as prevent that survey from being deleted. 
- Set the value of this column to correspond to the output of @bpasquer's query from the backlog issue.
- For locked surveys hide the select checkbox in the Survey List to prevent the survey from being added to a corrections job
- Also display the message "Survey data are locked" on the Survey Metadata Edit page
- Update unit tests.


utas-raymondng (changes requested):
Do I need to apply any sql change to my local DB? I got blank screen when I try to List Survey

nspool:
> Do I need to apply any sql change to my local DB? I got blank screen when I try to List Survey

@utas-raymondng Yes, first run the script db/schema-update/01-survey-lock.sql

utas-raymondng (approved):
LGTM

PR #1191, opened 2022-11-27
Display Survey Exists for M3 as warning

nspool:
Survey Exists was previously ignored for M3. Now display as a warning and skip the survey completion check if it fails.


utas-raymondng (approved):
LGTM

PR #1197, opened 2022-11-30
4512 improve performance

utas-raymondng:
Run a bit faster now, from the explain plan, the most costly part is still the string_agg (for diver and method)

->  GroupAggregate  (cost=484002.59..520016.13 rows=29870 width=68)
"                                Output: sm.survey_id, string_agg(DISTINCT (sm.method_id)::text, ', '::text ORDER BY (sm.method_id)::text), string_agg(DISTINCT (dr.full_name)::text, ', '::text ORDER BY (dr.full_name)::text)"
                                Group Key: sm.survey_id

No species by default and hidden until you expand the filter. Max two species filter allowed


nspool (approved):
Survey List is much faster now.

PR #1200, opened 2022-12-13
Shared Link Feature

nspool:
This PR contains the MVP for the Shared Link Feature, allowing users to create and expire copies of the private endpoint, currently limited to the EP_SITE_LIST endpoint. Not all of the endpoints have been implement as this is just a prototype.

- CSVs for each of the private endpoints are generated in code each morning at after the endpoints have been regenerated and stored in S3
- When a link is generated the file is copied to a public bucket with a unique URL
- URLs will be expired automatically (currently set to 1 day) or on demand


api/src/main/java/au/org/aodn/nrmn/restapi/service/upload/S3IO.java
```diff
@@ -1,41 +1,91 @@
...
+            var sessionId = RandomStringUtils.random(20, 0, 0, true, true, null, new SecureRandom());
+            var sourceKey = String.join("/", List.of("endpoints", viewName + ".csv"));
+            var destinationKey = String.join("/", List.of("endpoints", sessionId, viewName + ".csv"));
+            var request = CopyObjectRequest.builder()
+                    .sourceBucket(bucket).sourceKey(sourceKey)
+                    .destinationBucket(bucketShared).destinationKey(destinationKey)
+                    .acl(ObjectCannedACL.PUBLIC_READ)
+                    .expires(expires.toInstant(OffsetDateTime.now().getOffset()))
+                    .build();
+            client.copyObject(request);
```
utas-raymondng, line 61: Bug here should call getClient() instead of client otherwise null pointer exception for client, alternatively use @PostConstruct to init the client object is fine too

utas-raymondng: Bug: Create the share link with past expires date (for example today is 4 Jan, you create a link of 3 Jan) then copy the link and try to download it still works

utas-raymondng: Bug: Click the (x) in the expires field, it become dd/mm/yyyy, enter other info and click the create link will cause internal server error, I think it is better to either block user from click the button or give a better message instead of 500

utas-raymondng (changes requested):

utas-raymondng (approved):
LGTM

PR #1204, opened 2023-01-04
Fix M11 validation

nspool:
If M11 is allowed (ATRC corrections only) skip method check on M11 rows Otherwise validate based on species methods.


utas-raymondng (approved):
LGTM

PR #1207, opened 2023-01-05
Change Password Feature

nspool:
Adds a user menu that is displayed when the user email is clicked, with options to Change Password or Logout.
Change Password will change user password provided it is at least 8 characters and then immediately log out.
The `expires` field has been added to the `sec_user` table. When set to a date in the past the user will be redirected to the Change Password page and cannot proceed until the password has been updated.
Please ensure that `03-add-secuser-passwd-reset-flag.sql` has been applied before merging.


utas-raymondng:
The password update screen somehow prefilled the existing password for me, need a different variable name?

image

utas-raymondng (approved):
LGTM

PR #1208, opened 2023-01-10
Validate that survey IDs match corrected rows

nspool:
Adding or removing surveys in a corrections sheet should not be allowed.

- Survey Deletion should happen on the Survey Edit page
- Survey Creation should happen during Job Ingest.


utas-raymondng (approved):
LGTM

PR #1210, opened 2023-01-12
Add Shared Link CSVs + Improve UI

nspool:
- Implement `ep_m1` and `ep_m2_cryptic_fish` CSV exports for all, australia and states filters.
- Add the ability to generate more than one link at a time.
- Also improve CSS for Firefox and adds a Firefox debugging target for VSCode


utas-raymondng (approved):
Minor comment: You can generate the same endpoint with the same recipient and expiry on same date.

nspool:
@utas-raymondng will see if this or any other validations need to be implemented.

PR #1218, opened 2023-01-20
Add browser titles

nspool:
Every browser tab should now have the same title as the main header on the page. This makes opening multiple tabs much less confusing.


utas-raymondng (approved):
LGTM

PR #1221, opened 2023-01-30
MEOW Validation

nspool:


api/src/main/java/au/org/aodn/nrmn/restapi/data/repository/MeowRegionsRepository.java
```diff
@@ -1,9 +1,19 @@
...
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 
 import au.org.aodn.nrmn.restapi.data.model.MeowEcoRegions;
 
 public interface MeowRegionsRepository extends JpaRepository<MeowEcoRegions, Integer>, JpaSpecificationExecutor<MeowEcoRegions> {
+    
+    @Query(value = "WITH stp1 AS(" +
+        "SELECT DISTINCT(ecoregion)AS region, species_name FROM {h-schema}ep_m2_cryptic_fish GROUP BY species_name,ecoregion) " +
```
bpasquer, line 15: l can provide the query to account for M0,M1,M2cryptic,M2inverts,M7M8,M9 and M10

nspool: Thanks @bpasquer that would be much appreciated!

nspool: Actually @bpasquer is it sufficient to check that st_contains(meow.geom, site.geom) for any method? If so I'll just do that instead.

PR #1234, opened 2023-02-07
Improve L5/L95/Lmax validation messages

nspool:
Shows the size class in the message of each measurement validation.

- combine correction and ingestion validation grouping methods
- show description rows for each measurement along with the size class
- hide counts of 0, display count at end of summary
- show hover in summary as a dotted underline link


utas-raymondng (approved):
LGTM

PR #1235, opened 2023-02-08
Global Ingestion Lock

nspool:
- add global lock table to schema
- add global lock service
- wrap validation and ingest with lock
- remove all locks on application startup
- web: display an error if request fails

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/service/GlobalLockService.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/service/GlobalLockService.java
@@ -1,5 +1,7 @@
 package au.org.aodn.nrmn.restapi.service;
 
+import javax.transaction.Transactional;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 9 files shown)


api/src/main/java/au/org/aodn/nrmn/restapi/service/GlobalLockService.java
```diff
@@ -0,0 +1,30 @@
...
+
+import au.org.aodn.nrmn.restapi.data.model.GlobalLock;
+import au.org.aodn.nrmn.restapi.data.repository.GlobalLockRepository;
+
+@Service
+public class GlobalLockService {
+
+    @Autowired 
+    GlobalLockRepository globalLockRepository;
+
```
utas-raymondng, line 14: Need to use transaction around this section to make read consistency?

utas-raymondng (changes requested):
May need @Transcation to protect the lock

api/src/main/java/au/org/aodn/nrmn/restapi/service/GlobalLockService.java
nspool: The save in the GlobalLockService is where the lock is acquired, and this method is atomic.
I'll add a `@Transactional` to the setLock method, but in practice it shouldn't matter.

utas-raymondng (approved):
LGTM

PR #1241, opened 2023-02-15
Add geom indexes

nspool:
this appears to speed up the meow ecoregion check by a few seconds.


srw999 (approved):
Looks good!

nspool:
Thanks @srw999 !

PR #1245, opened 2023-02-28
Remove applied schema updates

nspool:
- Remove SQL files that have been applied to nrmn_prod
 - Add updated schema to application.sql
 - Add a useful script for creating an empty NRMN DB


utas-raymondng (approved):
LGTM

PR #1248, opened 2023-03-02
Server-side corrections sheet diff

nspool:
Diff the correction and original sheet on the server. Display the count of rows added and deleted.
For rows that have been changed display a list of edits.


utas-raymondng:
1. Correct this survey 923400688, do not make any change and hit submit. Is it ok to apply correction to survey without change?
2. I am not sure if it is my local db issue, please try to validate a few survey, I found the BLOCKING didn't looks right
image
3. Again may be my local db issue but correct this survey 923400686, make no change and then submit. Old value change from 0 to 0
image

bpasquer:
> 1. Is it ok to apply correction to survey without change?

That's a fair point. Maybe we should'not allow it.

nspool:
@utas-raymondng fixed the formatting of column 0 which should appear as Unsized. Not able to reproduce, but blocking errors may happen on historical surveys. Please see if you can reproduce against the `nrmn_dev` db on `db1-dev-nec-hob.emii.org.au`.

utas-raymondng:
> db1-dev-nec-hob.emii.org.au

Try this one 923401415

Connected to the nrmn_dev

1. Open 923401415, without changing anything, click validation, item highlighted but seems not correct 
image
2. Click submit correction, calculated total do not match with the highlight too 
image

utas-raymondng:
I will approve this for now as the error seems not related to this change.

utas-raymondng (approved):
LGTM

PR #1250, opened 2023-03-02
4710 fix error 404

utas-raymondng:


nspool:
@utas-raymondng I think it's an error 400, not 404.

nspool (changes requested):
If you encounter the constraint and error, return to the search and complete another correction the error message still shows. eg:

image

It would be good also if the error appears in an red outline alert to match other fatal messages.

nspool:
@utas-raymondng build appears to be failing due to unused vars.

utas-raymondng:
> @utas-raymondng build appears to be failing due to unused vars.

Yes, thanks for the remind. I am still in the mid of adding test case, so please ignore it for now.

PR #1254, opened 2023-03-08
Store bulk species update filters

nspool:
- Extend the StagedJobLog to include filters
- Also log the count of the observations changes
- Display the filters in a basic format if present on the job.


utas-raymondng (approved):
LGTM

PR #1257, opened 2023-03-09
Exclude SND auto-filled values from corrections diff

nspool:
Diver and Inverts are filled in by the UI and query respectively, since these values are not store for SND rows.


nspool:
Merging but please still review when you have time @utas-raymondng

PR #1260, opened 2023-03-17
Code and project improvements

nspool:
- Improve API Types by using DTOs instead of ad-hoc maps
- Move VSCode environment variables to `.env.development.local`


utas-raymondng (approved):
LGTM

PR #1266, opened 2023-03-27
User management functions

nspool:
For users with the Admin role this will display a list of users with the following functions:
- Add New User will create a new user with the ROLE_DATA_OFFICER and assign a random password that must be changed on first login.
- Edit will change the user email or full name.
- Disable Account will prevent the user from logging in next time but removing all user roles and setting account status to DEACTIVATED.
- Reset Password will assign a random password that must be changed on first login.
- Enable Admin will give user access to the admin options from the user menu (toggles the ROLE_ADMIN). Admin users will appear highlighted in yellow.

### DIFF
--- a/api/src/main/java/au/org/aodn/nrmn/restapi/controller/AdminController.java
+++ b/api/src/main/java/au/org/aodn/nrmn/restapi/controller/AdminController.java
@@ -4,22 +4,29 @@ import au.org.aodn.nrmn.restapi.data.model.SecRole;
 import au.org.aodn.nrmn.restapi.data.model.SecUser;
 import au.org.aodn.nrmn.restapi.data.repository.SecRoleRepository;
 import au.org.aodn.nrmn.restapi.data.repository.SecUserRepository;
+import au.org.aodn.nrmn.restapi.dto.auth.SecUserDto;
 import au.org.aodn.nrmn.restapi.enums.SecRoleName;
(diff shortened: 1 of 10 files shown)


utas-raymondng:
These warning appear in the console

WARNING in [eslint] 
src/components/admin/UserComponent.js
  Line 27:6:  React Hook useEffect has a missing dependency: 'onAdd'. Either include it or remove the dependency array. If 'onAdd' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps
  Line 31:6:  React Hook useEffect has a missing dependency: 'onAdd'. Either include it or remove the dependency array. If 'onAdd' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps

src/components/admin/UserList.js
  Line 62:30:  'i' is defined but never used  no-unused-vars

nspool:
@utas-raymondng fixed linter warnings. Also refactored into a reducer.

utas-raymondng:
> @utas-raymondng fixed linter warnings. Also refactored into a reducer.

Looks good now. A minor issue that I accidentally hit the "disable account" of the admin user of current login which ends up not able to do anything until I add the value back in the db. It may be better to show warning if user try to disable the current login user.

utas-raymondng (approved):
LGTM

PR #1268, opened 2023-03-29
Build artifact in a Docker container

nspool:
This workflow can be used as a template to migrate other Jenkins pipelines to GitHub Actions.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,42 +1,63 @@
-# This workflow builds the application and runs unit and integration tests
-
-name: Build and run tests
+name: Docker Build Pipeline
 
 on:
-#  - push
-  - pull_request
-
-concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/maven.yml
```diff
@@ -1,42 +1,63 @@
...
-name: Build and run tests
+name: Docker Build Pipeline
 
 on:
-#  - push
-  - pull_request
-
-concurrency:
-  group: ${{ github.ref }}
-  cancel-in-progress: true
```
digorgonzola, line 11: Don't we want to keep this?

digorgonzola (changes requested):

.github/workflows/maven.yml
nspool: Yes we do 😬
I'll add it back.

digorgonzola (approved):

PR #1269, opened 2023-04-04
Format measurement column names in survey diff

nspool:
Show the invert or fish size as the column name in the survey diff.


utas-raymondng (approved):
LGTM

PR #1272, opened 2023-04-21
Min width column to show applied filter

utas-raymondng:


web/src/components/import/DataSheetView.js
nspool, line 1: It might be better to change the minWidth in the defaultColDef instead of adding it to the columns.

utas-raymondng: Done

nspool (approved):

nspool:
LGTM

PR #1277, opened 2023-05-16
db update scripts: survey updates

bpasquer:


srw999 (approved):
Given there are lots of updates, it might be worthwhile to backup the table first.

PR #1278, opened 2023-05-18
4852 largefile timeout

utas-raymondng:


vietnguyengit:
do you have sample of a large file @utas-raymondng ?

vietnguyengit:
## Notes to be considered in a follow-up backlog item:

- Validation for "Total" seems not working probably (requiring either 0 or 1, but none of the values will pass).
- The "Submit" button is clickable although there are tons of blocking validation errors.

vietnguyengit:
- The loop of `jobStatus` checking requests (endpoint: `/api/v1/ingestion/ingest/<jobId>` ) works as expected for the case to expect an error.

api/src/main/java/au/org/aodn/nrmn/restapi/controller/IngestionController.java
```diff
@@ -25,84 +26,157 @@
...
+
+        result.put("jobStatus", stagedJobLog.getEventType());
+        result.put("jobLogId", stagedJobLog.getId());
+        return ResponseEntity.ok(result);
+    }
 
-        } catch (Exception e) {
+    @GetMapping(path = "ingest/{job_id}")
+    @Operation(security = { @SecurityRequirement(name = "bearer-key") })
+    public ResponseEntity<Map> getIngest(@PathVariable("job_id") Long jobId) {
```
vietnguyengit, line 165: `job_id` should be named `jobLogId` because this GET request takes `jobLogId` param instead of `jobId`

api/src/main/java/au/org/aodn/nrmn/restapi/controller/IngestionController.java
```diff
@@ -25,84 +26,157 @@
...
 
-            logger.error("Ingestion Failed", e);
+        Map<String, Object> result = new HashMap<>();
+        result.put("jobId", jobId);
```
vietnguyengit, line 168: same here, should be renamed to `jobLogId`.

vietnguyengit: for example, the POST request with `jobId` 205 as param returns 

```json
{"jobId":205,"jobStatus":"INGESTING","jobLogId":683}
```

this GET request takes `jobLogId` 683 as param but returns a confusing/incorrect `jobId` value:

```json
{"jobId":683,"jobStatus":"INGESTING","message":null}
```

api/src/main/java/au/org/aodn/nrmn/restapi/controller/IngestionController.java
```diff
@@ -25,84 +26,157 @@
...
-    UserActionAuditRepository userActionAuditRepository;
+    protected SpeciesFormattingService speciesFormatting;
 
     @Autowired
-    private SpeciesFormattingService speciesFormatting;
+    protected MaterializedViewService materializedViewService;
 
     @Autowired
-    private MaterializedViewService materializedViewService;
+    protected GlobalLockService globalLockService;
```
vietnguyengit, line 67: are there specific reasons that you changed to `protected`? I don't see where else other than the `IngestionController` class these variables are being used.

web/src/api/api.js
```diff
@@ -226,17 +228,38 @@ export const submitJobFile = (params, onProgress) => {
...
+    return;
+  };
+
+  let notDone = true;
+  let getResponse;
+
+  while(notDone) {
+    getResponse = await inst.get('ingestion/ingest/' + postResponse.jobLogId, { validateStatus: () => true }).then(res => res.data);
+    notDone = getResponse.jobStatus === 'INGESTING';
+    if(notDone) await sleep(timeout);
```
vietnguyengit, line 256: Does this delay 5 secs between every GET request checking ingestion status? This may slow down the whole process unnecessarily and require users to keep the browser tab open for a while.

What if they close the tab by accident or the internet drops out in the middle of the loop?

Are there any specific reasons for the timeout to be 5 seconds? It took me less than 20ms for the GET request to response.

image

vietnguyengit (changes requested):

api/src/main/java/au/org/aodn/nrmn/restapi/controller/IngestionController.java
```diff
@@ -25,84 +26,157 @@
...
-            var rows = rowRepository.findRowsByJobId(job.getId());
-            var species = speciesFormatting.getSpeciesForRows(rows);
-            var validatedRows = speciesFormatting.formatRowsWithSpecies(rows, species);
+        final var job = optionalJob.get();
+        if (job.getStatus() != StatusJobType.STAGED) {
+            result.put("jobStatus", StatusJobType.FAILED);
+            result.put("message", "Job with given id has not been validated: " + jobId);
+            logger.warn(result.get("message").toString());
+            return ResponseEntity.badRequest().body(result);
+        }
```
vietnguyengit, line 109: Can we also reject the request if a job has been validated but with `BLOCKING` validation errors?

vietnguyengit (changes requested):

web/src/api/api.js
vietnguyengit: This seems to be the most important to consider, the ingestion process running in the background now appears to be a long SYNC process simply because of waiting for GET request responses.

vietnguyengit: image

I'm having over 800 GET requests after nearly an hour, with no indicators of when this would finish and stuck in this screen:

image

vietnguyengit: My understanding is that the ingestion still happens in the background, potentially reaching "INGESTED" status at the end, and maybe failing too. How can we inform the user that feedback if the scenario above happens?

vietnguyengit: My opinion that asking users to wait for ingestion status check from the GET requests and seeing "waiting" on the frontend for sometime doesn't contribute to the whole ingestion process...

I'd prefer after clicking "submit" button, give them no more wait, instead telling them to come back after some time for the ingestion status. 

When they refresh the page or come back later to see the progress, in the job list page, show the status associated to their jobs.

web/src/api/api.js
```diff
@@ -257,4 +280,4 @@ export const templateZip = (params) => {
 
 export const originalJobFile = (jobId) => {
   return axiosInstance.get(`stage/job/download/${jobId}`, {responseType: 'blob'});
-};
+};
```
vietnguyengit, line 283: still miss end line

vietnguyengit:
> ## Notes to be considered in a follow-up backlog item:
> * Validation for "Total" seems not working probably (requiring either 0 or 1, but none of the values will pass).
>   image
>   image
>   image
>   I deleted the troublesome rows above to get through for the purposes of reviewing this PR.
> * The "Submit" button is clickable although there are tons of blocking validation errors.
> * With too many "survey exists" rows, I wonder if can we have an effective way to bulk remove them.
>   image
>   I deleted all surveys and their other foreign key constraints to get through this for the purposes of reviewing this PR, the list is way too much to manually delete one by one.
>   image
>   no blocking now.

Can you please open to follow-up ticket for two first two items as bugs? The last one is a feature, by the way, it's up to you to add it or not.

vietnguyengit (approved):
LGTM, front-end bugs will be investigated in separated backlog item.

Issue #1280, opened 2023-06-05
species_list not displaying some species


bpasquer:
From email 2023-06-03:
Just quickly, I have noticed an issue with the ep_species_list and subsequently the species_list_data views. For example, Apogon doederleini (observable_item_id 810) which has been superseded with Ostorhinchus doederleini (observable_item_id 8114) is not displaying in either endpoint. For some reason 810 is present in nrmn.observation, but 8114 isn’t, and because 810 has been superseded by 8114 – neither are being displayed in ep_species_list (or species_list_data on the geoserver). This is not a recent superseding, but we have noticed this change recently in the website species pages (i.e. there is now no record of this species). Has something changed in the way the species_list is being generated?

This issue doesn’t seem to be consistent, and I can’t really see why? The only thing I noticed was that Ostorhinchus doederleini has been imported into the Observable Items table from WoRMS and WoRMS has done another taxonomic reshuffle whereby all Phylum/Class/Orders seem to have changed for some species. I have manually changed the Phylum/Order/Class of Ostorhinchus doederleini back to be consistent with Apogon doederleini, and will check tomorrow if the species appears in the species_list layer. If it does (and the higher taxonomy is the cause of the species omission from the species_list endpoints) then I think we need to look at working out a periodical WoRMS scrape/update for these changes and apply them to relevant species consistently.

bpasquer:
The reason why _Apogon doederleini_  and  _Ostorhinchus doederleini_  are missing from ep_species_list is two-fold and explained by the condition in the code to create ep_species_list:
```sql
where oi.obs_item_type_name in ('Species', 'Undescribed Species')
and exists (select 1 from nrmn.observation obs where obs.observable_item_id = oi.observable_item_id)
and oi.superseded_by is null)
```

So:
 - superseded species are excluded from ep_species_list, hence _Apogon doederleini_ is excluded
 - ep_species_list contains species that have at least one observation, and because _Ostorhinchus doederleini_  has no attached observation, it is excluded too

The condition should check whether a species has ever been observed based on its _current_ name, not on the observable_item_id

atcooper1:
Yes, @bpasquer the condition should be using current name not observable_item_id. It looks like all recent superseding corrections are not being reflected in the Reef Species of the World pages because of this. 

Supersedings are displaying correctly/as expected in data endpoints.

bpasquer:
actually the condition 
```sql 
and oi.superseded_by is null
```
in 
```sql
[...]

PR #1283, opened 2023-06-14
remove condition in ep_species_list to exclude superseded species

bpasquer:
resolves https://github.com/aodn/nrmn-application/issues/1280


utas-raymondng (approved):
LGTM

PR #1289, opened 2023-06-19
pg_cron sql scripts for nrmn

srw999:
need somewhere to put the pg_cron sql code for NRMN 
see [convert existing cron jobs to pg_cron](https://github.com/aodn/backlog/issues/4881)


utas-raymondng (approved):
LGTM

PR #1299, opened 2023-07-11
Fix/4962 delete supersededby

utas-raymondng:


web/src/components/input/CustomSearchInput.js
```diff
@@ -64,8 +64,10 @@ const CustomSearchInput = ({label, exclude, formData, onChange, fullWidth}) => {
...
     }
     else {
-      // Programmatic change or 'clear'
-      if(textValue !== value) {
+      // Programmatic change or 'clear' || run check on init case
+      // it is used to handle case where the species no longer exist but
+      // used in supersededby.
+      if(textValue !== value || searchTerm === null) {
         // Update if diff to avoid looping
         setSearchTerm(value);
```
vietnguyengit, line 72: Can you please clarify "clear" in this context? If it's when `bird` is removed before `eagle` and expecting `superseded by` field value when editing `eagle` to be emptied, then it doesn't work.
image

and also by reading :

> it is used to handle case where the species no longer

without previous context knowledge, I assume line 72 would be reached only when the description above happens. However, it seems when root species still exist, the line still be reached. e.g `sample` is parent of `pigeon`

image

and when `bird` is removed while it was parent of `eagle`

image

other than that, the main issue which was a blank page is gone.

utas-raymondng: > Can you please clarify "clear" in this context? If it's when `bird` is removed before `eagle` and expecting `superseded by` field value when editing `eagle` to be emptied, then it doesn't work. image
> 
> and also by reading :
> 
> > it is used to handle case where the species no longer
> 
> without previous context knowledge, I assume line 72 would be reached only when the description above happens. However, it seems when root species still exist, the line still be reached. e.g `sample` is parent of `pigeon`
> 
> image
> 
> and when `bird` is removed while it was parent of `eagle`
> 
> image
> 
> other than that, the main issue which was a blank page is gone.

Refer to this https://mui.com/material-ui/api/autocomplete oninputchange, i think it means text box clear.

The code reach is after the value set in the field, it will trigger a lookup in the database and verify if the mentioned species is a valid in db. So if you delete the species, it will prompt for error.

vietnguyengit: cool thanks @utas-raymondng for explaining

PR #1308, opened 2023-08-21
Bug/5021 seperate speciesfamily screen

utas-raymondng:


vietnguyengit:
reviewing this PR

web/src/common/relation-graph/FamilyTree.js
```diff
@@ -255,6 +263,7 @@ FamilyTree.propTypes = {
   focusNodeId: PropTypes.number,
   onSkipLengthWeightChange: PropTypes.func,
   onSaveLengthWeightChange: PropTypes.func,
+  onExistEdit: PropTypes.func,
```
vietnguyengit, line 266: is that a typo with `exist`, e.g `exit` instead? other than that, the rest of LGTM.

I noted that, for species that don't have length-weight values set at the root. e.g

image

to be able to use the Family tree editing modal, I had to click `Save observable item` after putting the values into the field, otherwise, the Family tree editing modal will be empty. But I think this is how the users will be happy with, and the root species also not in the scope of the new button `Superseded(by) species`

the feature works as expected

image


if it's a type, please update any occurrence, if not, I'm happy to merge this PR.

vietnguyengit (dismissed):

vietnguyengit (dismissed):

vietnguyengit (approved):

vietnguyengit (approved):

vietnguyengit (dismissed):
LGTM

PR #1324, opened 2023-10-30
Bug in using lower precision on double value

utas-raymondng:


vietnguyengit (approved):

bpasquer:
@utas-raymondng the decision in https://github.com/aodn/backlog/issues/3032 was to store lat and lon in float consistently across the api and database, not the opposite as the aim is to provide the coordinate with 5 decimal.

PR #1332, opened 2023-11-30
fix the bug (total=0 rows) and the corresponding test case.

HavierD:
Now, duplicate rows with total=0 in the middle of the sheet won't be removed. Only bottom duplicate rows with total=0 will be removed.


api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
```diff
@@ -230,4 +230,13 @@ void duplicateRowsShouldBeRemoved() throws Exception {
         assertEquals(20, validRows.size());
     }
 
+    @Test
```
utas-raymondng, line 233: Can you add comment to your test case so that what it is doing?

utas-raymondng (changes requested):

api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
HavierD: > Can you add comment to your test case so that what it is doing?

Got it.

utas-raymondng:
Don't merge until nrmn software version upgrade completed

api/src/main/java/au/org/aodn/nrmn/restapi/controller/StagedJobController.java
```diff
@@ -378,4 +381,25 @@ public void downloadFile(final HttpServletResponse response, @PathVariable Strin
                     jobId, clientException.getMessage())));
         }
     }
+
```
utas-raymondng, line 384: Bad function call that modify pass in object value, please return something instead of modify the existing one

Collection<Long> filterNonBottomDuplicateRows(Collection<Long>..... , Long);

api/src/main/java/au/org/aodn/nrmn/restapi/controller/StagedJobController.java
```diff
@@ -378,4 +381,25 @@ public void downloadFile(final HttpServletResponse response, @PathVariable Strin
...
+    private void filterNonBottomDuplicateRows(Collection<Long> rowIdsToRemove, Long lastRowId) {
+
+        var toRemoveIds = new ArrayList<Long>();
+
+        if (!rowIdsToRemove.contains(lastRowId)) {
```
utas-raymondng, line 389: Is this a reliable check, can the lastRowId not empty and then it will not exist in the rowIdsToRemove()?

api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
```diff
@@ -230,4 +230,18 @@ void duplicateRowsShouldBeRemoved() throws Exception {
...
 
+    /** This test is for:
+     * When adding new jobs, only rows (with "total=0" and duplicate more than 3 times) at the end of the sheet should be
+     * removed. Rows (with "total=0" and duplicate more than 3 times) are not at the end of the sheet should be kept.
+     * @throws Exception
+     */
+    @Test
+    void removeBottomDuplicateRowsOnly() throws Exception {
+        FileSystemResource file = new FileSystemResource("src/test/resources/sheets/removeDuplicateBottomRows.xlsx");
+        MockMultipartFile mockFile = new MockMultipartFile("sheets/removeDuplicateBottomRows.xlsx", file.getInputStream());
```
utas-raymondng, line 241: The removeDuplicateBottomRows contains 495 lines, the last 7 line is total=0 and repeat more then 3 times,  the first two line are header, so 495 - 7 - 2 = 486 so why 487 is expected result?

api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
```diff
@@ -230,4 +230,18 @@ void duplicateRowsShouldBeRemoved() throws Exception {
...
+        var parsedSheet = sheetService.stageXlsxFile(mockFile, false);
+        var validRows = stagedJobController.getRowsToSave(parsedSheet);
+        assertEquals(487, validRows.size());
+    }
```
utas-raymondng, line 245: Please add a few edge case, 

1. The xls file last line is "ID = 486"
2. The xls file last line is "ID = 490"
3. Assert shows that line "ID = 464, 465, 466, 467, 468" exist in the validRows

utas-raymondng (changes requested):
Please see comments

api/src/main/java/au/org/aodn/nrmn/restapi/controller/StagedJobController.java
HavierD: Got it. Will modify soon.

api/src/main/java/au/org/aodn/nrmn/restapi/controller/StagedJobController.java
HavierD: Sorry I didn't get your point. The meaning of this part is: Because we only want to remove duplicate rows on the bottom. So, if the `rowIdsToRemove` doesn't contain the last row, it means no rows should be removed.

api/src/test/java/au/org/aodn/nrmn/restapi/service/SpreadSheetServiceIT.java
HavierD: to record after discussion: if the last row is valid (species is "SURVEY NOT DONE", "DEBRIS - ZERO", or "NO SPECIES FOUND"), it will keep this row. Need to add a new test for it (valid, keep it. Otherwise remove it).

utas-raymondng (approved):
LGTM

PR #1335, opened 2023-12-07
Fix issue with software version upgrade on Jest test

utas-raymondng:


utas-raymondng:
1. Goto Observable Item click the pen icon show error sheetManager undefined
2. New Observable item also shows the same error
3. Site also have similar problem on edit

HavierD:
Yes I think it is because of the removed makeStyles library. We are using React 18 but makeStyles(from @mui/styles) only support React untill version 17. 
image

I think this pull request should be blocked until issue [#5224](https://github.com/aodn/backlog/issues/5224) has solved.

utas-raymondng:
Close as replaced by other PR

PR #1336, opened 2023-12-13
Species template script

bpasquer:


utas-raymondng (approved):
LGTM as user confirmed result

utas-raymondng:
Fix with https://github.com/aodn/nrmn-application/pull/1339

PR #1340, opened 2024-01-03
 Select species by current name and total abundance

bpasquer:


utas-raymondng:
@bpasquer I have rollback the change and look at the original sql in the getAllWithMethodForSites, there is a clause to skip item having supersededBy

"where m.method_id = :methodId AND site_raw.site_id IN (:siteIds) AND obsitem.superseded_by IS NULL "

bpasquer:
> "where m.method_id = :methodId AND site_raw.site_id IN (:siteIds) AND obsitem.superseded_by IS NULL " hence the template should never show supersededBy entries.

Yes this clause is removing the superseded species, but what we are trying to achieve here is list superseding species with no observation attached, at that can only be achieved through  their related superseded species

utas-raymondng:
@bpasquer I make couple of changes, and here is the template generated from my local PC, do you think the result correct?

image

[template(5).zip](https://github.com/aodn/nrmn-application/files/13855436/template.5.zip)

bpasquer:
> @bpasquer I made couple of changes, and here is the template generated from my local PC, do you think the result correct?
> 
> image
> 
> [template(5).zip](https://github.com/aodn/nrmn-application/files/13855436/template.5.zip)

@utas-raymondng   the template has  the expected entries :+1:

utas-raymondng:
@bpasquer so this PR contains your change right?

utas-raymondng (approved):
LGTM

utas-raymondng:
deployed to systest with Version 12.3 (854)

PR #1341, opened 2024-01-05
Upgrade/replace usage of @mui styles #5224

HavierD:


HavierD:
this should be merged after the merging of package version updating.

PR #1344, opened 2024-03-21
replace usage of @mui styles version2#5224

HavierD:
Remove all usage of mui/styles, and use "styled" and 'sx' to instead.

### DIFF
--- a/web/src/components/data-entities/survey/SpeciesCorrectEdit.js
+++ b/web/src/components/data-entities/survey/SpeciesCorrectEdit.js
@@ -1,5 +1,7 @@
 import React, {useCallback, useEffect} from 'react';
 
+import { styled } from '@mui/material/styles';
+
 import {Button, Box, LinearProgress} from '@mui/material';
 
 import IconButton from '@mui/material/IconButton';
@@ -19,15 +21,21 @@ import CustomSearchFilterInput from '../../input/CustomSearchFilterInput';
 import {searchSpecies, postSpeciesCorrection} from '../../../api/api';
 import {Paper} from '@mui/material';
 
-import {makeStyles} from '@mui/styles';
+const PREFIX = 'SpeciesCorrectEdit';
 
-const useStyles = makeStyles(({palette, typography}) => ({
-  root: {
+const classes = {
+  root: `${PREFIX}-root`
+};
+
+const StyledBox = styled(Box)(({
+                                 theme: {palette, typography}
+                               }) => ({
+  [`& .${classes.root}`]: {
     '& .MuiTable-root': {
       '& .MuiTableHead-root': {
         '& .MuiTableRow-head': {
           '& .MuiTableCell-head': {
-            fontSize: typography?.table.fontSize,
+            fontSize: typography?.table?.fontSize,
             background: palette?.primary.rowHeader
           }
         }
@@ -38,8 +46,8 @@ const useStyles = makeStyles(({palette, typography}) => ({
         }
       },
       '& .MuiTableCell-root': {
-        fontSize: typography?.table.fontSize,
-        padding: typography?.table.padding,
+        fontSize: typography?.table?.fontSize,
+        padding: typography?.table?.padding,
         color: palette?.text.textPrimary
       }
     }
@@ -54,7 +62,7 @@ const SpeciesCorrectEdit = ({selected, onSubmit, onError}) => {
     invertCondition: false,
     surveyIds: []
   };
-  const classes = useStyles();
+
   const pageSize = 10;
   const [page, setPage] = useState(0);
   const [detail, setDetail] = useState();
@@ -67,16 +75,16 @@ const SpeciesCorrectEdit = ({selected, onSubmit, onError}) => {
       setLoading(true);
       searchSpecies(selected.filter).then((res) => {
         const detail = res.data
-          .map((v) => ({locationName: v.locationName, siteName: v.siteName, surveyIds: JSON.parse(v.surveyIds)}))
-          .reduce((p, v) => {
-            const location = p.find((l) => l.locationName === v.locationName);
-            if (location) {
-              location.sites.push({siteName: v.siteName, surveyIds: v.surveyIds});
-            } else {
-              p.push({locationName: v.locationName, sites: [{siteName: v.siteName, surveyIds: v.surveyIds}]});
-            }
-            return p;
-          }, []);
+            .map((v) => ({locationName: v.locationName, siteName: v.siteName, surveyIds: JSON.parse(v.surveyIds)}))
+            .reduce((p, v) => {
+              const location = p.find((l) => l.locationName === v.locationName);
+              if (location) {
+                location.sites.push({siteName: v.siteName, surveyIds: v.surveyIds});
+              } else {
+                p.push({locationName: v.locationName, sites: [{siteName: v.siteName, surveyIds: v.surveyIds}]});
+              }
+              return p;
+            }, []);
         setDetail(detail);
       });
     }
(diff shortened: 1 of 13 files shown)


utas-raymondng:
I found these errrors

1. Click "List Surveys" -> select a survey -> click "Correct Survey Data" go back to front page.
2. The checkbox disappear on the page, please refer to edge on the same page.

utas-raymondng (changes requested):
I found these errors:

1. Click "List Surveys" -> select a survey -> click "Correct Survey Data" go back to front page.
2. The checkbox disappear on the page, please refer to edge on the same page.

HavierD:
> I found these errors:
> 
> Click "List Surveys" -> select a survey -> click "Correct Survey Data" go back to front page.
> The checkbox disappear on the page, please refer to edge on the same page.

For recording: 
I think this error is just because, in our local environment, the `app.features` is empty (in the dege/systest envirnoment. the app.features are added while deploying). We can just add `corrections` into app.features ( like this: `app.features=corrections`) for temporarily testing. After this, the two errors will all go. The app.features is in: api.src.main.resources.application.properties

utas-raymondng (approved):
LGTM

HavierD:
Final check approved from me.

web/src/components/data-entities/survey/SpeciesCorrectEdit.js
```diff
@@ -19,15 +21,21 @@ import CustomSearchFilterInput from '../../input/CustomSearchFilterInput';
 import {searchSpecies, postSpeciesCorrection} from '../../../api/api';
 import {Paper} from '@mui/material';
 
-import {makeStyles} from '@mui/styles';
+const PREFIX = 'SpeciesCorrectEdit';
+
+const classes = {
+  root: `${PREFIX}-root`
+};
```
vietnguyengit, line 28:
```javascript
const PREFIX = 'SomeName';
const classes = {
  root: `${PREFIX}-root`
};
```

this block appears the same in multiple places, is there any specific reason it cannot be just

```javascript
const classes = {
  root: `SomeName-root`
};
```
even so, I'd refer to have list of constants grouped in a separate file to import later 

e.g

create a file `constant.js`

```javascript
const SOME_NAME_CLASSES = {
  root: "SomeName-root"
};

const ANOTHER_NAME_CLASSES = {
  root: "AnotherName-root"
};

export {
  SOME_NAME_CLASSES,
  ANOTHER_NAME_CLASSES
};
```

then in the component you need one of the above constant, you can import it with

```javascript
import { SOME_NAME_CLASSES as classes } from "./path/to/constants.js";
```

vietnguyengit: @HavierD

vietnguyengit: having 

```javascript
const PREFIX = ...;
const classes = ...;
```
spread everywhere may not be ideal. please change.

HavierD: Thank you Viet. I will check it later after finishing my current ticket.

HavierD: @vietnguyengit I remember the reason. After researching the usage of `@mui/material/styles`, it must have a unique class name otherwise it will have CSS class collisions. At the beginning, my code is similar with this: 
```javascript
const classes= {
  root: "SomeName-root"
};
```

But some components have multiple style class, i have to be like this: 
```javascript
const classes= {
  root: "SomeName-root",
  header:"SomeName-header",
  body":"SomeName-body"
...
};
```

So I think, extracting the prefix of the classes is a good idea.

And for consistence and maintenance, i extract all the prefixes if the `styled()` from `@mui/material/styles` is in use, even though some of them only have one class. 

To make sure the prefix is unique, i use the component name as the prefix (I know that occasionally the name might be duplicated, but it is easy to solve and for the current nrmn, this is enough.).  
So currently, all the components using `styled()` have this structure (which is neat and easy to understand and maintain I reckon):

```javascript
const PREFIX = <component_name>
const classes = {
  <name1>: `${PREFIX}-name1`,
  <name2>: `${PREFIX}-name2`
 ...
}

```
And because the PREFIX is just the same name with the component, so i think putting PREFIX into the component file rather than putting all PREFIXs into one `constraint.js` is better.

The above is my understanding and explanation. Please let me know if something is incorrect.

vietnguyengit:
```javascript
const PREFIX = <component_name>
const classes = {
  <name1>: `${PREFIX}-name1`,
  <name2>: `${PREFIX}-name2`
 ...
}
```

which could looklike

```javascript
const SOME_NAME_CLASSES = {
  root: "SomeName-root"
};

const ANOTHER_NAME_CLASSES = {
  root: "AnotherName-root"
};

const NEW_CLASSES = {
  <name1>: "Component-name1",
  <name2>: "Component-name2"
 ...
}

export {
  SOME_NAME_CLASSES,
  ANOTHER_NAME_CLASSES,
  NEW_CLASSES
};
```

you still can declare the constants from dedicated/centralised place for all the constants, my point is you introduced duplicate code structure everywhere unnecessarily.

Please make new PR to maintain good code practice. @HavierD

vietnguyengit: @HavierD  https://semaphoreci.com/blog/constants-layer-javascript#:~:text=such%20a%20layer!-,3%20Approaches%20for%20Organizing%20Constants%20in%20a%20JavaScript%20Project,-Let%E2%80%99s%20see%20three

vietnguyengit: different language but still the same point : https://medium.com/swlh/why-you-should-use-a-constants-file-in-swift-ff8c40af1b39

vietnguyengit: https://dev.to/amirfakour/tips-to-use-constants-file-in-typescript-27je

HavierD: I know your point. If several files share one PREFIX & classes, it is a good idea to put them together into a constant.js (like some normal constants such as names, URLs, theme parameters etc.). But the PREFIXes i am using are privately used. They can only be used in one file, and must not be shared (will cause css class collision). Putting them together will make them be public rather than private. As we all know, make private members public is dangerous.

vietnguyengit: > I know your point. If several files share one PREFIX & classes, it is a good idea to put them together into a constant.js (like some normal constants such as names, URLs, theme parameters etc.). But the PREFIXes i am using are privately used. They can only be used in one file, and must not be shared (will cause css class collision). Putting them together will make them be public rather than private. As we all know, make private members public is dangerous.

if you are sure they wont be used anywhere else, and like you mentioned that the whole point you tried to do it is for having unique names as required by your new UI dependency, then again, you still litter the code with duplicate structure everywhere

```javascript
const PREFIX = <component_name>
const classes = {
  <name1>: `${PREFIX}-name1`,
  <name2>: `${PREFIX}-name2`
 ...
}
```

please make new PR to avoid using duplicate duplicate and hard-coded `PREFIX` approach, you can use **dynamic** & unique string approach for example, because like you said, nowhere else uses them,  @HavierD

vietnguyengit: even you don't use the exported const anywhere, they are still constants 

>  make private members public is dangerous.

i dont think you understand my feedback and the sample articles? @HavierD

vietnguyengit: again, please make new PRs to improve your code standard to avoid using duplicate code structure unnecessarily. I moved the ticket to `In progress` and put it as a blocker until you make improvements. @HavierD

web/src/components/import/custom/header.js
```diff
@@ -1,31 +1,41 @@
...
-import {makeStyles} from '@mui/styles';
 
-const useStyles = makeStyles((theme) => ({
-  fishSize: {
+const PREFIX = 'AgGridHeader';
+
+const classes = {
+  fishSize: `${PREFIX}-fishSize`,
+  invertSize: `${PREFIX}-invertSize`
+};
```
vietnguyengit, line 10: I still think you need to work out a way to clean this,

```javascript
const classes = {
  fishSize: `${PREFIX}-fishSize`,
  invertSize: `${PREFIX}-invertSize`
};
```

the same thing appears in `GridHeader.js`

```javascript
const classes = {
  fishSize: `${PREFIX}-fishSize`,
  invertSize: `${PREFIX}-invertSize`
};
```

just like


```javascript
const classes = {
  root: `${PREFIX}-root`
};
```

appears in `SpeciesCorrectEdit.js` `SpeciesCorrectErrorResults.js` `SpeciesCorrectResults.js` and `SurveyDiff.js`

```javascript
const classes = {
  root: `${PREFIX}-root`
};
```

```javascript
const classes = {
  root: `${PREFIX}-root`
};
```

```javascript
const classes = {
  root: `${PREFIX}-root`
};
```

```javascript
const classes = {
  root: `${PREFIX}-root`
};
```

----

above are duplicate codes from your PR, which you can clearly able to identify a syntax

```javascript
const classes = {
  <element>: <unique-name>,
  ...
};
```

Too many duplicate codes to be acceptable, maybe a helper function for generate unique class names for different components if the class name generator from MUI not fit? https://mui.com/material-ui/experimental-api/classname-generator/

if you do go with a helper function let's say something like this should reduce a lot of duplicate code @HavierD 

```javascript
// Define a helper function to generate unique class names
function generateUniqueClasses(prefix, classNames) {
  const generatedClasses = {};
  
  for (const className of classNames) {
    generatedClasses[className] = `${prefix}-${className}`;
  }

  return generatedClasses;
}

// Usage example
const PREFIX = "yourPrefix";

const aClasses = generateUniqueClasses(PREFIX, ["fishSize", "invertSize"]);
// OR
const bClasses = generateUniqueClasses(PREFIX, ["root"]);

console.log(aClasses);
console.log(bClasses);
```

HavierD: Your solution has such side effects:

- Helper is good for extracting a complicated block, and use them with much simple ways. But in this case, it only change this: 
- 
```javascript

const classes = {root: `${PREFIX}-root`};

```

      into this:

```javascript

const bClasses = generateUniqueClasses(PREFIX, ["root"]);

```

        and also with an extra helper class.


- introducing an external library only for several string constants. NRMN now doesn't use uuid. Introducing it directly is not a good idea if it can only be replaced by several string declaration. 
 https://medium.com/@hello.shahanshah/problems-of-using-too-many-libraries-and-how-to-avoid-them-dccf9ac8e029
 https://www.linkedin.com/pulse/best-practices-using-javascript-libraries-frameworks-dibor-alfred/

-  uuid generates meaningless long random characters such as 104u1340qijfeaor324nj, but declaring PREFIX directly is meaningful. For example, if someone want to debug using DevTools Elements, the corresponding css class is like this: 
image
which is much readable and maintainable than random characters like: 104u1340qijfeaor324nj-root. No one can easily know which component this meaningless-character CSS class is in relation to. 

- Split the same portion of code. Currently, the code is like this: 
```javascript
const PREFIX = 'GridHeader';

const classes = {
  fishSize: `${PREFIX}-fishSize`,
  invertSize: `${PREFIX}-invertSize`
};

const Root = styled('div')(({theme}) => ({
  [`& .${classes.fishSize}`]: {
    color: '#c4d79b',
    borderBottom: '1px solid ' + theme.palette.divider
  },

  [`& .${classes.invertSize}`]: {
    color: '#da9694'
  }
}));

```
Which is very readable and no comments needed. Developers are easy to know: first variable is taking the component name as the prefix, second object is defining the css class names using the right above prefix, and the third object is defining the css values using the classes right above as a part of selectors. 

Thank you for your concern, but I think, in all the comments, my solution is better than others (as i explained above). Besides, not all similar structure needs to be extracted, and not all constants need to be assembled into a constant file. Using meaningless class name to instead meaningful class name is not good code practice; exposing a private member (which is dangerous to be public) is even worse code practice, etc. I accept all the recommendations which is better than mine (as I did in a lot of tickets), but currently i don't think this ticket needs to change, otherwise @utas-raymondng wouldn't approve my PR. I don't think my code of this ticket is bad and worth so much time to explain, I won't create a new pull request. If you still insist on your point and think my explanation is useless, please ask @vfisaac and @utas-raymondng for help to review all the points of us in this page, and make final decision.

vietnguyengit: You have 2 samples, one with uuid and one not. Whatever ways, they are my suggestions to improve your way of duplicating code, I'm not fixing the PR for you with actual implementation, you need to workout and brainstorm a final way to improve it, but not duplicating code like you are using.

I will not satisfied with a bunch of duplicated lines with the reasons you mentioned.

I leave this up to the bosses @utas-raymondng @vfisaac

HavierD: They are not the common duplicating code as you reckoned. They are just one string declaration and one object declaration, and they must be private and these two are a part of the CSS block for a component. Private members with similar structure almost occurs everywhere in any projects. So I don't think I need to workout anyway to improve it.

vietnguyengit: > They are not the common duplicating code as you reckoned. They are just one string declaration and one object declaration, and they must be private and these two are a part of the CSS block for a component. Private members with similar structure almost occurs everywhere in any projects. So I don't think I need to workout anyway to improve it.

The helper function itself can be public while the data it generates or processes remains private within the component where it's used. Your code are unnecessarily duplicated.

vietnguyengit: Hi @HavierD, As an outcome of the meeting with other dev, please link this PR to either your new improvement PR or backlog ticket that you will work on to refactor the code, so the issue here is traceable for future reference. Thank you.

HavierD: Thank you for the reminding. I will do it shortly.

HavierD:
After the meeting, I am open to accept others' opinions and reserve my opinion. The extracted variables have been inlined. The pr is: https://github.com/aodn/nrmn-application/pull/1349

PR #1345, opened 2024-03-22
libraries upgrading.

HavierD:


Dockerfile
```diff
@@ -1,24 +1,42 @@
-FROM amazoncorretto:11-al2-jdk
+FROM ubuntu:22.04
```
utas-raymondng, line 1: May I know why ubuntu is in use instead of a smaller image of amazoncorretto11?

Dockerfile.build
```diff
@@ -1,22 +1,40 @@
-FROM amazoncorretto:11-al2-jdk
+FROM ubuntu:22.04
```
utas-raymondng, line 1: Same as above

web/package.json
```diff
@@ -11,18 +11,20 @@
     "@mui/lab": "^5.0.0-alpha.84",
     "@mui/material": "5.13.2",
     "@mui/styles": "5.13.2",
```
utas-raymondng, line 13: Remove this line @mul/styles as I hit issue when remove the node_modules directory and npm install

https://github.com/mui/material-ui/issues/32982

And then a few more place need to fix once this dropped

utas-raymondng (changes requested):
As per comments

Dockerfile
vietnguyengit: I know the answer to this question:

`amazoncorretto` image based on Linux2 which doesn't support the installation of Node 18's dependency called `glibc 2.28`; hence cannot use `node18` on `amazoncorretto` image:

ref:

- https://repost.aws/questions/QUvkkhY--uTiSDkS6R1jFnZQ/node-js-18-on-amazon-linux-2
> Unfortunately Amazon Linux2 does not support Nodejs18.

- https://serverfault.com/questions/1132783/how-can-i-get-node18-on-amazonlinux2
> You cannot run the binary on your system because it relies on a version of glibc which is not available on your system.

understanding there is config for v18 in `package.json` but that is for React not Node

`"react": "18.1.0"`

if do not want to upgrade node engine then can just simply use `amazoncoretto` image; but there are benefits of using Node 18 to be considered:

https://www.mhtsai.me/blog/top-5-differences-between-node-16-and-node-18/

web/package.json
HavierD: I think it works now. Could you help me check it again?

utas-raymondng (approved):
LGTM

PR #1347, opened 2024-04-08
Bug/lat&lon rounding issue#5275

HavierD:
- Change the truncating method into rounding.
- Add a validate method: if the lat & lon in the added job is within 10 metres, make them "null".


api/src/main/java/au/org/aodn/nrmn/restapi/service/validation/SiteValidation.java
```diff
@@ -92,6 +92,13 @@ public SurveyValidationError validateSurveyAtSite(StagedRowFormatted row) {
                 return new SurveyValidationError(ValidationCategory.DATA, ValidationLevel.WARNING, message,
                         Arrays.asList(row.getId()), Arrays.asList("latitude", "longitude"));
             }
+            // If survey is less than 10 meters, warn users survey lat & lon will be removed.
```
bpasquer, line 95: @HavierD I don't think this warning is necessary. However, there should be a warning when survey coordinates  are outside of a 10m radius  from site coords , and the warning should display the difference in metres

HavierD: @bpasquer Got it. Will change it shortly.

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
```diff
@@ -100,7 +101,7 @@ public Survey getSurvey(Program program, OptionalDouble visAvg, StagedRowFormatt
 
         Optional<Survey> existingSurvey = surveyRepository.findOne(Example.of(survey));
 
-        return existingSurvey.orElseGet(() -> surveyRepository.save(
+        var returningSurvey = existingSurvey.orElseGet(() -> surveyRepository.save(
```
utas-raymondng, line 104: You already do a save here with incorrect data, because surveyRepository.save () called.

Can you move the distanceBetween before this line and then the function call become  

.longitude(distance < 10d ? null : stagedRow.getLongitude())

utas-raymondng (changes requested):
Small change request

api/src/main/java/au/org/aodn/nrmn/restapi/service/SurveyIngestionService.java
HavierD: oh I miss that small `save` word and reckoned that part is just to build a survey. Thank you for reminding.

HavierD: Will try to do some slight change about the format.

HavierD: @utas-raymondng Hello Raymond. I have fixed that problem. Could you help me check the PR again?

api/src/main/java/au/org/aodn/nrmn/restapi/service/upload/SurveyContentsHandler.java
```diff
@@ -131,13 +131,20 @@ public void cell(String cellReference, String formattedValue, XSSFComment commen
...
     }
 
-    private String truncateDecimalString(String decimalString) {
-        int idx = decimalString.indexOf('.');
-        int endIdx = idx + 6;
-        if (idx >= 0 && endIdx < decimalString.length())
-            return decimalString.substring(0, endIdx);
-        else
+
+    private static String roundDecimalString(String decimalString) {
```
utas-raymondng, line 135: Sorry forgot to comment this:

is this shorter with common-math (already included in pom) -> Precision.round(numnber, 5);

By default, it is using the same HALF_UP rounding method as our helper method; therefore, the results should be the same.

then format with 
DecimalFormat df = new DecimalFormat("###.###");
System.out.println(df.format(PI));

https://www.baeldung.com/java-round-decimal-number

utas-raymondng (changes requested):
Other ok,  just one comment

api/src/main/java/au/org/aodn/nrmn/restapi/service/upload/SurveyContentsHandler.java
utas-raymondng: One thing also puzzle me is that before the change, the save() should save the incorrect value, why the auto test not picking it up?

utas-raymondng (approved):
LGTM

PR #1364, opened 2024-05-30
Feature/edit data with warning#5599

HavierD:
In the system, now:

- in the staged job sheets (list jobs -> staged), lat & lon are rounded but not empty (if within 10m of the site).
- In the ingested job sheets (list jobs -> ingested), lat & lon are rounded but not empty (if within 10m of the site) even though users edited the sheet after staging.
- In Database, the lat & lon of the Survey table are rounded or null (if within 10m of the site).


utas-raymondng:
So the solution is to change the save action, right now we will round the lat/lng automatically before save and then do the validation, therefore there is not way to generate a warning on the UI.

This change split the save to two one, the first save do not round it automatically, then give user a chance to see the warning where lat/lng will be round

When submit, those lat/lng will be round automatically and then save it to db and then call submit.

utas-raymondng (approved):
LGTM , further test needed in edge as I cannot fully test in local due to s3 permission issue

Issue #1368, opened 2024-06-19
Some surveys in survey_list have incorrect contents in 'method' field


LizziOh:
For example survey

912351271 and 912351272 - neither have method 2 observations however have 1,2 in the methods field in the survey_list endpoint.

![image](https://github.com/aodn/nrmn-application/assets/52521522/e2cdde0c-7249-4011-b373-8805a1a9c50c)

They display correctly in the UI

![image](https://github.com/aodn/nrmn-application/assets/52521522/1c8aef24-2b29-4b77-b59c-1d8210a1eb22)

@bpasquer @atcooper1

LizziOh:
I believe its an issue that when 'Survey not done' is TRUE then that method gets recorded as done.

bpasquer:
Hi Lizzi, you are almost right:
As you said,  M2 is correctly recorded as 'survey_not_done' in the survey_method table: 
| survey_method_id | survey_id | method_id | block_num | survey_not_done |
|---|---|---|---|---|
| 12485 | 912351271 | 2 | 1 | true |
| 12486 | 912351271 | 1 | 1 | false |
| 12487 | 912351271 | 1 | 2 | false |
| 12488 | 912351271 | 2 | 2 | true |
| 56987 | 912351272 | 2 | 2 | true |
| 56988 | 912351272 | 1 | 1 | false |
| 56989 | 912351272 | 2 | 1 | true |
| 56990 | 912351272 | 1 | 2 | false |

This code snippet of the ep_survey_list endpoint provides the list of methods done in a survey:

```sql
SELECT string_agg(DISTINCT sm1.method_id::character varying(3)::text, ', '::text
                          ORDER BY (sm1.method_id::character varying(3)::text)) AS string_agg
        FROM nrmn.survey_method sm1
        WHERE sm1.survey_id = sur.survey_id
        GROUP BY sm1.survey_id
```
As you can see, it simply aggregates all the methods listed in the survey_method table above without excluding records where survey_not_done is True. Adding a condition as below will produce the correct output:

```sql 
SELECT string_agg(DISTINCT sm1.method_id::character varying(3)::text, ', '::text
                         ORDER BY (sm1.method_id::character varying(3)::text)) AS string_agg
        FROM nrmn.survey_method sm1
        WHERE sm1.survey_id = sur.survey_id AND sm1.survey_not_done=false
        GROUP BY sm1.survey_id
```

LizziOh:
Thanks for looking into this Bene, 
[...]

PR #1369, opened 2024-06-20
exclude survey not done from method list in survey endpoint

bpasquer:
adresses https://github.com/aodn/nrmn-application/issues/1368


utas-raymondng (changes requested):
I try to run the sql with the test case and the below failed, so looks like this change have more cascade impact

[ERROR]   ATRCSurveyGroupCompleteIT.groupWithIncompleteSurveyBlocksShouldFail:242
[ERROR]   RLSMethodBlockAssociationIT.missingOnBlockShouldFail:118 expected: <true> but was: <false>
[ERROR]   MaterializedViewServiceIT.verifyEpSurveyList:97 Site list first match ==> array contents differ at index [15], expected: <null> but was: <Jan Jansen, Liz Oh>
[ERROR]   WormsServiceIT.partialSearchReturnResultEqualsThanFifty:117 Record size match expected:<50> but was:<0>
[ERROR]   WormsServiceIT.partialSearchReturnResultLessThanFifty:99 Record size match expected:<14> but was:<0>
[ERROR]   WormsServiceIT.partialSearchReturnResultMoreThanFifty:135 Record size match expected:<50> but was:<0>

utas-raymondng (approved):

PR #1374, opened 2024-07-04
Data update

bpasquer:
These are the scripts to address the following tasks:
1)  both the Fishbase update to latest version (2023-05)+ rounding of values ( a's to 5 decimals, b's to 2 decimals). Two sripts are provided , one for doing the update of existing values, the other to insert new values. These two can be combined into one when applied.  
2) a- Rounding of survey coordinates 
   b- deletion of coordinate when the survey was done at the site location


utas-raymondng (approved):
LGTM

PR #1386, opened 2024-10-14
Feature/5778 add test before speed improve

utas-raymondng:
There is a need to improve processing speed, add test case before major change


HavierD (approved):
test cases added only. looks good

PR #1399, opened 2025-04-14
Features/6118 speedup save validation

utas-raymondng:
1. Simplify syntax
2. The getAll on Site is use to create a Map, so it is better to create the map in the query
3. findSiteCodesMatching is very expensive as it load Site object internally but all we need to Site code, so replace it with nativeQuery


HavierD (approved):
LGTM

PR #1418, opened 2026-02-17
RRH embargoed data

bpasquer:


HavierD (approved):
ran in edge, systest and prod. LGTM

PR #1427, opened 2026-04-22
SQL scripts for WoRMS update

bpasquer:
The PR contains two scripts that can be applied in any order as they are unrelated.
Aphia_Insert.sql:  
- deletes the content of table aphia_ref
- (re)populates the table. 
This table is currently static (ie, no process in the NRMN system is interacting with this table).This script will result in the update of existing entries as well as addition of new entries in the table.

Taxonomy_update.sql
- updates the aphia_id, phylum,order,class,family in observable_item_ref


HavierD:
Taxonomy_Update.sql alrady run in both edge and systest

HavierD:
@bpasquer hello is it ready to run in prod?

bpasquer:
@HavierD no, the query needs adjustements

shaunahu:
I tried running `Aphia_Insert.sql` locally, but it reported a few issues that are likely to also fail in edge/systest/prod.

First, `Aphia_Insert.sql` contains duplicate `aphia_id` values:

```sql
ERROR: duplicate key value violates unique constraint "aphia_ref_pkey"
Key (aphia_id)=(205965) already exists.

SQL state: 23505
Detail: Key (aphia_id)=(205965) already exists.
```

The following `aphia_id` values appear to be duplicated across multiple rows, for example in rows starting with `VALUES ('205362'`:

* `205362`
* `205965`
* `204781`
* `142241`
* `106840`
* `144132`

After I removed the duplicated rows, I encountered another error:

```sql
ERROR: value too long for type character varying(200)

SQL state: 22001
```

The affected fields are:

* Line 5729: `unacceptreason` has a limit of 200 characters, but the value length is 251.

  * Value starts with: `the reinstatement of the laevis specimen as type species calls for resurrection of the original genus name. In spite of`
* Line 21161: `authority` has a limit of 100 characters, but the value length is 104.

  * Value: `Korshunova, Fletcher, Picton, Lundin, Kashio, N. Sanamyan, K. Sanamyan, Padula, Schrödl & Martynov, 2020`
* Line 21161: `valid_authority` has a limit of 100 characters, but the value length is 104.

  * Value: `Korshunova, Fletcher, Picton, Lundin, Kashio, N. Sanamyan, K. Sanamyan, Padula, Schrödl & Martynov, 2020`

@bpasquer could you please check these values before we apply the script to all environments?

Remaining tasks:

* [ ] Apply `Taxonomy_Update.sql` in production
* [ ] Apply `Aphia_Insert.sql` in edge, systest and production

PR #1432, opened 2026-07-07
re-estabish fields in public endpoints

bpasquer:


shaunahu:
both sql have been applied in edge and systest

shaunahu:
applied in prod

shaunahu (approved):

PR #1434, opened 2026-07-16
position of program restored in private endpoint

bpasquer:


shaunahu (approved):

shaunahu:
have applied in edge, systest, and prod

PR #1435, opened 2026-07-17
Feature/1431 publication endpoints to s3

shaunahu:
Function updated in this PR (I am using the `nrmn-dev` bucket in account `nonprod` for local dev, further app config is required for cloudformation):
1. extract public NRMN endpoint data as per csv files
image

2. add TPAC and RLS endpoints in scheduled task
image

3. Set up a copy in the _shared bucket, using a common obfuscation key across the three RLS endpoints.
image

4. Set up a copy to _shared bucket with a obfuscation key for the TPAC endpoint
image


api/src/main/java/au/org/aodn/nrmn/restapi/service/PublicViewService.java
```diff
@@ -0,0 +1,117 @@
...
+                logger.info("Published public view " + relation + "_data");
+            } catch (Exception e) {
+                logger.error("Failed to publish public view " + relation, e);
+            }
+        }
+        stopWatch.stop();
+        logger.info("Published all public views in " + stopWatch.getLastTaskTimeMillis() + "ms");
+    }
+
+    private void publishView(String relation, String outputName) throws Exception {
```
utas-raymondng: It would be better to use the Repository Object similar to this https://github.com/aodn/nrmn-application/blob/main/api/src/main/java/au/org/aodn/nrmn/restapi/data/repository/DiverRepository.java so that we have a single place for SQL and no need to do those transaction with code

utas-raymondng (changes requested):
Request change to SQL

utas-raymondng (approved):
LGTM

